PR #1, opened 2022-04-19
Feature stream any im instance logs

tonsV2:


pkg/instance/handler.go
```diff
@@ -466,7 +468,7 @@ func (h Handler) Logs(c *gin.Context) {
 		_ = c.Error(err)
 	}
 
-	readCloser, err := h.instanceService.Logs(instance, group)
+	readCloser, err := h.instanceService.Logs(instance, group, selector)
```
teleivo, line 471: I would rename `readCloser` to `r` and any writer to `w` as this is how its usually done in the standard library and other Go projects.

swagger/swagger.yaml
```diff
@@ -382,6 +382,12 @@ paths:
       description: Stream instance logs in real time
       operationId: instanceLogs
       parameters:
+      - description: 'name: selector'
+        format: uint64
```
teleivo, line 386: should format and type both be string?

pkg/instance/service.go
```diff
@@ -168,23 +171,26 @@ func (s service) Logs(instance *model.Instance, group *models.Group) (io.ReadClo
 	return read, err
 }
 
-func (s service) getPod(client *kubernetes.Clientset, instance *model.Instance) v1.Pod {
+func (s service) getPod(client *kubernetes.Clientset, instance *model.Instance, selector string) (v1.Pod, error) {
+	labelSelector := fmt.Sprintf("im-id=%d", instance.ID)
```
teleivo, line 175: I do really like Go's guard clause/reduce nesting style. In this case this should go into an else statement, as its a waste to compute a labelSelector if there is a non empty `selector`.

pkg/instance/service.go
```diff
@@ -168,23 +171,26 @@ func (s service) Logs(instance *model.Instance, group *models.Group) (io.ReadClo
 	return read, err
 }
 
-func (s service) getPod(client *kubernetes.Clientset, instance *model.Instance) v1.Pod {
+func (s service) getPod(client *kubernetes.Clientset, instance *model.Instance, selector string) (v1.Pod, error) {
```
teleivo, line 174: Can you add a guard (returning an error) here if selector is anything other than `""` or `"data"` like we talked about. So we are reminded of the connection to stacks (don't accidentally allow streaming of `im-data-secret-id`, ...).

teleivo (commented):
I think the swagger.yml needs to be changed. Other suggestions are minor.

teleivo (approved):

PR #16, opened 2022-05-18
Feat/use parameter name as pk

tonsV2:


pkg/instance/handler.go
```diff
@@ -104,7 +104,7 @@ func (h Handler) Create(c *gin.Context) {
 }
 
 type ParameterRequest struct {
-	StackParameterID uint   `json:"stackParameterId" binding:"required"`
+	StackParameterID string `json:"stackParameterId" binding:"required"`
```
teleivo, line 107: Can/should we rename this to `StackParameter` (at some point)?

pkg/instance/handler.go
```diff
@@ -192,29 +192,31 @@ func (h Handler) Deploy(c *gin.Context) {
 	c.JSON(http.StatusCreated, instance)
 }
 
-func convertRequiredParameters(instanceID uint, requestParameters []ParameterRequest) []model.InstanceRequiredParameter {
+func convertRequiredParameters(instance *model.Instance, requestParameters []ParameterRequest) []model.InstanceRequiredParameter {
 	if len(requestParameters) > 0 {
 		var parameters = make([]model.InstanceRequiredParameter, len(requestParameters))
 		for i, parameter := range requestParameters {
+			stackRequiredParameter := model.StackRequiredParameter{ID: parameter.StackParameterID}
```
teleivo, line 199: I would inline this (and in function below) into the struct literal if not used anywhere else. saves us naming/reading/maintaining a var in our head 😋

pkg/instance/handler.go
```diff
@@ -192,29 +192,31 @@ func (h Handler) Deploy(c *gin.Context) {
 	c.JSON(http.StatusCreated, instance)
 }
 
-func convertRequiredParameters(instanceID uint, requestParameters []ParameterRequest) []model.InstanceRequiredParameter {
+func convertRequiredParameters(instance *model.Instance, requestParameters []ParameterRequest) []model.InstanceRequiredParameter {
```
teleivo, line 195: Why did you change the parameter from uint to *model.Instance? All you need in here is the uint, right?

pkg/instance/helmfile.go
```diff
@@ -95,12 +95,12 @@ func (h helmfileService) configureInstanceEnvironment(accessToken string, instan
...
 
 	for _, parameter := range instance.RequiredParameters {
-		instanceEnv := fmt.Sprintf("%s=%s", parameter.StackRequiredParameter.Name, parameter.Value)
+		instanceEnv := fmt.Sprintf("%s=%s", parameter.StackRequiredParameter.ID, parameter.Value)
 		cmd.Env = append(cmd.Env, instanceEnv)
 	}
 
 	for _, parameter := range instance.OptionalParameters {
-		instanceEnv := fmt.Sprintf("%s=%s", parameter.StackOptionalParameter.Name, parameter.Value)
+		instanceEnv := fmt.Sprintf("%s=%s", parameter.StackOptionalParameter.ID, parameter.Value)
```
teleivo, line 103: why rename it from Name to ID?

pkg/stack/loader.go
```diff
@@ -77,7 +77,7 @@ func LoadStacks(stackService Service) {
 }
 
 func extractRequiredParameters(file []byte) []string {
-	regexStr := "\\{\\{[ ]?requiredEnv[ ]?\"(.*?)\"[ ]?\\}\\}"
+	regexStr := "\\{\\{[ ]?requiredEnv[ ]?\"(.*?)\".*\\}\\}"
```
teleivo, line 80: does that relate to switching from ids to names?

scripts/deploy-job.sh
```diff
@@ -15,26 +15,7 @@ PAYLOAD='{\"databaseId\": \"11\"}'
 echo "{
   \"requiredParameters\": [
     {
-      \"stackParameterId\": 5,
-      \"value\": \"$JOB\"
-    }
-  ],
-  \"optionalParameters\": [
-    {
-      \"stackParameterId\": 10,
```
teleivo, line 24: why did you delete the optional params?

scripts/dhis2-core-deploy.sh
```diff
@@ -26,21 +19,21 @@ echo "{
   \"stackId\": 1,
   \"optionalParameters\": [
     {
-      \"stackParameterId\": $READINESS_PROBE_INITIAL_DELAY_SECONDS_PARAMETER_ID,
+      \"stackParameterId\": \"READINESS_PROBE_INITIAL_DELAY_SECONDS\",
```
teleivo, line 22: 🎉 nice

scripts/dhis2-deploy.sh
```diff
@@ -2,23 +2,13 @@
...
-
-IMAGE_TAG="2.36.0-tomcat-8.5.34-jre8-alpine"
-IMAGE_TAG_PARAMETER_ID=1
-
-DATABASE_ID=1
-DATABASE_ID_PARAMETER_ID=1
+IMAGE_TAG=2.36.0-tomcat-8.5.34-jre8-alpine
+DATABASE_SIZE=30Gi
+PGADMIN_INSTALL=false
+DATABASE_ID=3
```
teleivo, line 11: DATABASE_ID changed from 1 to 3, why?

stacks/dhis2-core/helmfile.yaml
```diff
@@ -27,7 +27,7 @@ releases:
       - dhisConfig: |
           connection.dialect = org.hibernate.dialect.PostgreSQLDialect
           connection.driver_class = org.postgresql.Driver
-          connection.url = jdbc:postgresql://{{ requiredEnv "DATABASE_HOSTNAME" }}/{{ env "DATABASE_DATABASE" | default "dhis2" }}
+          connection.url = jdbc:postgresql://{{ requiredEnv "INSTANCE_NAME" }}-database-postgresql/{{ env "DATABASE_DATABASE" | default "dhis2" }}
```
teleivo, line 30: should this be changed? if so in the dhis2-core-deploy script DATABASE_HOSTNAME is still provided as required parameter

teleivo (commented):
I've added some questions. I'll need a bit more time to look at the pkg/model/stack.go GORM stuff. That is a bit hard to decipher 😂

pkg/instance/handler.go
tonsV2: Done!

scripts/deploy-job.sh
tonsV2: I actually didn't. The script, by mistake, had two echo commands and only the second piped its output to the http command.

It should have been addressed in a separate commit but if you don't object I'll include it here rather than reverting it.

scripts/dhis2-core-deploy.sh
tonsV2: Damn right!

pkg/stack/loader.go
teleivo: then what is it 😋 its easier to review a PR if it only contains whats in the PR title/description/task.

teleivo (approved):
can't wait to use the scripts in my environment 🎉 🥇

PR #17, opened 2022-05-20
chore: use updated rabbitmq consumer DEVOPS-130

teleivo:
* added integration test to show instance is deleted when message is received

if you break the implementation by for example changing the queue name to some other name you will see

```
    ttlDestroyConsumer_test.go:42: FAIL: 0 out of 1 expectation(s) were met.
                The code you are testing needs to make 1 more call(s).
                at: [ttlDestroyConsumer_test.go:42 asm_amd64.s:1571]
    ttlDestroyConsumer_test.go:41:
                Error Trace:    ttlDestroyConsumer_test.go:41
                Error:          Condition never satisfied
                Test:           TestSuiteTTLDestroyConsumer/TestConsumeDeletesInstance

```

Note: the failed expectations will be printed multiple times due to a retry mechanism needed since things happen asynchronously.

* if I reuse test helper for setting up a queue/client a 3rd time in the im-inspector, I will extract it into its own repository.


teleivo:
Opened a draft @tonsV2 so you can see how it would look when used 😄

tonsV2 (approved):

PR #30, opened 2022-06-11
refactor: create engine without wire

teleivo:


cmd/serve/main.go
```diff
@@ -42,25 +45,44 @@ func main() {
 }
 
 func run() error {
-	environment := di.GetEnvironment()
+	config := config.ProvideConfig()
```
tonsV2, line 48: Intellij will warn about variable name will collide with package name. Shall we agree on either cfg or just c? Whatever you prefer :-)

tonsV2 (approved):

cmd/serve/main.go
teleivo: ah, good catch! yes, I lean towards `cfg`. I think `c` is often used for a `Client` like struct for example in a receiver.

PR #31, opened 2022-06-14
feat: use service account when deleting an instance as a result of co…

tonsV2:
…nsuming an event otherwise use the actual user


pkg/instance/ttlDestroyConsumer.go
```diff
@@ -4,17 +4,22 @@ import (
...
 type ttlDestroyConsumer struct {
+	config          config.Config
+	userClient      userClient.Client
 	consumer        *rabbitmq.Consumer
 	instanceService Service
 }
 
-func ProvideTtlDestroyConsumer(consumer *rabbitmq.Consumer, instanceService Service) *ttlDestroyConsumer {
-	return &ttlDestroyConsumer{consumer, instanceService}
+func ProvideTtlDestroyConsumer(config config.Config, userClient userClient.Client, consumer *rabbitmq.Consumer, instanceService Service) *ttlDestroyConsumer {
```
teleivo, line 21: I would only pass in username and password as that's the only thing that is needed from the config.

pkg/instance/ttlDestroyConsumer_test.go
```diff
@@ -17,17 +19,20 @@ import (
...
 	consumer, err := rabbitmq.NewConsumer(
 		s.rabbitURI,
 		rabbitmq.WithConsumerPrefix("im-manager"),
 	)
 	require.NoError(err)
 	defer func() { require.NoError(consumer.Close()) }()
 
-	is := &instanceService{}
-	is.On("Delete", uint(1)).Return(nil)
+	is := &instanceService{userClient: uc}
```
teleivo, line 32: the instanceService does not depend on the userClient so you should not add it as a field.

pkg/instance/ttlDestroyConsumer_test.go
```diff
@@ -17,17 +19,20 @@ import (
...
+	is.On("Delete", "token", uint(1)).Return(nil)
```
teleivo, line 33: since you added a call to `userClient.SignIn(c.config.UserService.Username, c.config.UserService.Password)` in the implementation of Consume()

you will want to check that we actually signin using the correct credentials.

so I think something like

`uc.SignIn("user", "pw").Return("some token")`

pkg/instance/ttlDestroyConsumer_test.go
```diff
@@ -40,12 +45,38 @@ func (s *ttlSuite) TestConsumeDeletesInstance() {
...
+func (u *userClient) FindUserById(token string, id uint) (*models.User, error) {
+	return nil, nil
+}
+
+func (u *userClient) FindGroupByName(token string, name string) (*models.Group, error) {
+	return nil, nil
+}
+
+func (u *userClient) SignIn(username, password string) (*models.Tokens, error) {
+	return &models.Tokens{
```
teleivo, line 61: this will then need to change to something like the `Delete` implementation below

pkg/instance/ttlDestroyConsumer_test.go
tonsV2: Do you mean...

	uc.On("SignIn", "username", "password").Return(&models.Tokens{
		AccessToken: "token",
	}, nil)

?

teleivo (approved):

PR #50, opened 2022-07-03
feat: restart instances

tonsV2:
This PR allows us to restart instances

The following limitations exists
* Restart only works for deployments, not statefulsets and others
* The deployment needs to have the "app.kubernetes.io/instance" label
* Currently we can only restart whoami-go and dhis2-core instances (not pgadmin or postgresql databases, one is lacking the required label and the other is controlled by a statefulset and not a deployment)

### DIFF
--- a/pkg/instance/service.go
+++ b/pkg/instance/service.go
@@ -67,12 +67,10 @@ func (s service) Restart(token string, id uint) error {
 	}
 
 	err = s.kubernetesService.Executor(group.ClusterConfiguration, func(client *kubernetes.Clientset) error {
-		labelSelector := fmt.Sprintf("app.kubernetes.io/instance=%s", instance.Name)
-		listOptions := metav1.ListOptions{
-			LabelSelector: labelSelector,
-		}
-
 		deployments := client.AppsV1().Deployments(instance.GroupName)
+
+		labelSelector := fmt.Sprintf("app.kubernetes.io/instance=%s", instance.Name)
+		listOptions := metav1.ListOptions{LabelSelector: labelSelector}
 		deploymentList, err := deployments.List(context.TODO(), listOptions)
 		if err != nil {
 			return err
--- a/pkg/instance/service.go
+++ b/pkg/instance/service.go
@@ -78,11 +78,10 @@ func (s service) Restart(token string, id uint) error {
 
 		items := deploymentList.Items
 		if len(items) > 1 {
-			return fmt.Errorf("multiple deployments found using the selector: %s", labelSelector)
+			return fmt.Errorf("multiple deployments found using the selector: %q", labelSelector)
 		}
-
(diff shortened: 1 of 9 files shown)


pkg/instance/service.go
```diff
@@ -54,6 +55,72 @@ type userClientService interface {
...
+		deploymentList, err := deployments.List(context.TODO(), listOptions)
+		if err != nil {
+			return err
+		}
+
+		items := deploymentList.Items
+		if len(items) > 1 {
+			return fmt.Errorf("multiple deployments found using the selector: %s", labelSelector)
+		}
+
```
teleivo, line 85: I would remove the newline. The two ifs are related so can be grouped together

pkg/instance/ttlDestroyConsumer_test.go
```diff
@@ -60,6 +60,10 @@ type instanceService struct {
 	mock.Mock
 }
 
+func (is *instanceService) Restart(token string, id uint) error {
```
teleivo, line 63: We should isolate the ttlDestroyConsumer from such changes, since its actually only interested in the ability to destroy an instance.

Don't remember if you already have another PR somewhere to adapt the interface 😄

teleivo (commented):
small comments, questions 😄

pkg/instance/service.go
```diff
@@ -54,6 +55,72 @@ type userClientService interface {
...
+	}
+
+	group, err := s.userClient.FindGroupByName(token, instance.GroupName)
+	if err != nil {
+		return err
+	}
+
+	err = s.kubernetesService.Executor(group.ClusterConfiguration, func(client *kubernetes.Clientset) error {
+		labelSelector := fmt.Sprintf("app.kubernetes.io/instance=%s", instance.Name)
+		listOptions := metav1.ListOptions{
```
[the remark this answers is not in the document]
tonsV2: I feel it's packing a lot information into a single line which is why I use the variable. Please consider my updated semi-inlining

scripts/restart.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+INSTANCE_NAME=$1
+GROUP_NAME=$2
+
+INSTANCE_ID=$($HTTP --check-status "$INSTANCE_HOST/instances-name-to-id/$GROUP_NAME/$INSTANCE_NAME" "Authorization: Bearer $ACCESS_TOKEN")
```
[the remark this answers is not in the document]
tonsV2: Yes

pkg/instance/ttlDestroyConsumer_test.go
tonsV2: Nope, that was for the api gateway

PR #51, opened 2022-07-04
Feat/link stacks

tonsV2:
# Description

This PR is rather big as it includes...

* Implementation of the pgAdmin stack
* Simplification of deploy arguments (only a thin wrapper is used for the instance parameter)
* Changes to the stack parsing code
* A new type of stack parameters currently referred to as "consumables" to indicate they are consumed from another instance

It might make sense that we dedicate a meet to discuss what's been implemented. But the essence of it all is that upon deployment some of the instance parameters are extracted from another instance. The current use case is that a DHIS2 stack, or a pgAdmin stack, can extract parameters (such as username, password and hostname) from the dhis2-db stack.
But it could also be other kind of stacks "linking" against any kind stack exposing an endpoint. Such as a RabbitMQ, a Kafka or a MariaDB stack.

# Test

## Create database instance

    export MYID=tons-db-1 && export ACCESS_TOKEN="" && eval $(./login.sh) && echo $ACCESS_TOKEN && ./dhis2-db-create.sh $MYID whoami && ./dhis2-db-deploy.sh $MYID whoami && read && eval $(./login.sh) && echo $ACCESS_TOKEN && ./destroy.sh $MYID whoami

## Link pgAdmin instance against the database instance

    export MYID=tons-pgadmin-1 && export ACCESS_TOKEN="" && eval $(./login.sh) && echo $ACCESS_TOKEN && ./pgadmin-create.sh $MYID whoami && ./pgadmin-link.sh tons-db-1 $MYID whoami && read && eval $(./login.sh) && echo $ACCESS_TOKEN && ./destroy.sh $MYID whoami

## Link DHIS2 core instance against the database instance

    export MYID=tons-core-1 && export ACCESS_TOKEN="" && eval $(./login.sh) && echo $ACCESS_TOKEN && ./dhis2-core-create.sh $MYID whoami && ./dhis2-core-link.sh tons-db-1 $MYID whoami && read && eval $(./login.sh) && echo $ACCESS_TOKEN && ./destroy.sh $MYID whoami


pkg/instance/repository.go
```diff
@@ -24,6 +30,48 @@ func NewRepository(DB *gorm.DB) *repository {
...
+	link := &model.Linked{
+		FirstInstanceID:  firstInstance.ID,
+		StackName:        secondInstance.StackName,
+		SecondInstanceID: secondInstance.ID,
+	}
+	err := r.db.Create(&link).Error
+	if err != nil {
+		var perr *pgconn.PgError
+		if errors.As(err, &perr) && perr.Code == "23505" {
+			return fmt.Errorf("instance (%d) already linked with a stack of type \"%s\"", firstInstance.ID, secondInstance.StackName)
```
teleivo, line 43: use verb `%q` https://pkg.go.dev/fmt to quote the StackName 😄

pkg/instance/repository.go
```diff
@@ -24,6 +30,48 @@ func NewRepository(DB *gorm.DB) *repository {
...
+		return err
+	}
+	return nil
+}
+
+func (r repository) Unlink(instance *model.Instance) error {
+	link := &model.Linked{}
+
+	// Does another instance depend on the instance we're trying to unlink
+	err := r.db.First(link, "first_instance_id = ?", instance.ID).Error
```
teleivo, line 54: So `first_instance_id` would become `target_instance_id` in my understanding

pkg/instance/repository.go
```diff
@@ -24,6 +30,48 @@ func NewRepository(DB *gorm.DB) *repository {
...
+		return fmt.Errorf("instance %d depends on %d", link.SecondInstanceID, instance.ID)
+	}
+
+	// Any error beside ErrRecordNotFound?
+	if err != nil && !errors.Is(err, gorm.ErrRecordNotFound) {
+		return err
+	}
+
+	// Attempt to unlink
+	err = r.db.Unscoped().Delete(link, "second_instance_id = ?", instance.ID).Error
```
teleivo, line 65: and `second_instance_id` would become `source_instance_id`

pkg/instance/repository.go
```diff
@@ -24,6 +30,48 @@ func NewRepository(DB *gorm.DB) *repository {
...
+func (r repository) Link(firstInstance *model.Instance, secondInstance *model.Instance) error {
+	link := &model.Linked{
+		FirstInstanceID:  firstInstance.ID,
+		StackName:        secondInstance.StackName,
+		SecondInstanceID: secondInstance.ID,
+	}
+	err := r.db.Create(&link).Error
+	if err != nil {
+		var perr *pgconn.PgError
+		if errors.As(err, &perr) && perr.Code == "23505" {
```
teleivo, line 42: nice that you applied the error handling we introduced in the im-user 😄

pkg/stack/repository.go
```diff
@@ -33,6 +34,8 @@ func (r repository) Delete(name string) error {
 func (r repository) Find(name string) (*model.Stack, error) {
 	var stack *model.Stack
 	err := r.db.
+		//		Preload("RequiredParameters", "consumed <> ?", true).
```
teleivo, line 37: remove the commented out lines of code 🧹 😄

scripts/pgadmin-link.sh
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+PGADMIN_USERNAME=someone@something.com
+PGADMIN_PASSWORD=QYfzfZTeF8pVMwECvt7jgBdk
```
teleivo, line 6: Should we use something like https://unix.stackexchange.com/questions/230673/how-to-generate-a-random-string instead. Otherwise anyone can login to our pgadmins.

pkg/stack/loader.go
```diff
@@ -57,21 +60,41 @@ func LoadStacks(stackService Service) {
 			log.Fatalf("Failed to read stack: %s", err)
 		}
 
-		stackParameters := extractStackParameters(file)
+		hostnamePattern := extractMetadataParameters(file, hostnamePatternIdentifier)
+		if 0 < len(hostnamePattern) && len(hostnamePattern) < 2 {
```
teleivo, line 64: I think you mean `len(hostnamePattern) == 1`, right? If so please simplify to that expression 😄 

Should we return an error if `len(hostnamePattern) > 1` ?

pkg/stack/loader.go
```diff
@@ -57,21 +60,41 @@ func LoadStacks(stackService Service) {
...
+			stack.HostnamePattern = hostnamePattern[0]
+		}
+
+		hostnameVariable := extractMetadataParameters(file, hostnameVariableIdentifier)
+		if 0 < len(hostnameVariable) && len(hostnameVariable) < 2 {
```
teleivo, line 69: same as above

pkg/instance/service.go
```diff
@@ -54,6 +64,107 @@ type userClientService interface {
...
+	newStack, err := s.stackService.Find(newInstance.StackName)
+	if err != nil {
+		return err
+	}
+
+	// Consumed required parameters
+	for _, parameter := range newStack.RequiredParameters {
+		if parameter.Consumed && parameter.Name != newStack.HostnameVariable {
+			value, err := s.findParameterValue(parameter.Name, oldInstance, oldStack)
+			log.Println(parameter.Name, value)
```
teleivo, line 87: Could this print a plain text password? If so remove the line as we should not log such information.

I would even remove it regardless actually. To me such logs are debug logs that might only highlight that we need some tests.

pkg/stack/loader.go
```diff
@@ -57,21 +60,41 @@ func LoadStacks(stackService Service) {
...
 			}
 		}
 
 		optionalParameterMap := extractOptionalParameters(file, stackParameters)
 		fmt.Printf("Optional parameters: %+v\n", optionalParameterMap)
-		for k, v := range optionalParameterMap {
-			_, err := stackService.CreateOptionalParameter(stack, k, v)
+		for name, v := range optionalParameterMap {
+			isConsumed := isConsumedParameter(name, consumedParameters)
+			log.Println("name: ", name, "v: ", v, "isConsumed: ", isConsumed)
```
teleivo, line 96: I would remove this line. Similar to the other log line I commented on: to me such logs are debug logs that might only highlight that we need some tests.

teleivo (changes requested):
I would take a moment to think about naming with regards to linking before merging this.

We talked about it already but only relating `first/second`. During the review I found variables names for instances prefixed with old/new. It's not clear to me which one is old/new 😬. As far as I understand if I want to link pg-admin to dhis2-db I would need to create dhis2-db as well as pg-admin before calling the new `POST /instances/:id/link/:newInstancesId` endpoint. The pg-admin instance could even be older with regards to the creation date. It might not be with regards to the deploy date as it has not been deployed yet AFAIK. Its still not obvious.

If we were to settle on `source` and `target` we should use this throughout (swagger path parameters, function parameters, variables, fields, ...).

Would you agree that if I link pg-admin to dhis2-db that pg-admin is the source and dhis2-db is the target?

In that case if my understanding of the code is correct 😅 any names prefixed with

`first/old...` would become `target...`
`second/new...` would become `source...`

pkg/stack/loader.go
tonsV2: That is indeed what I mean! Doh! :roll_eyes: 

Yes, we should! However I think for now we should follow the pattern implemented in the rest of the loader and simply use log.Fatalf. After gaining more experience with Go I'm aware it's not ideal. My original thought was that this is happens at boot time and by using log.Fatalf I wanted to ensure it failed early if the stacks couldn't be parsed.

scripts/pgadmin-link.sh
teleivo: My point is not to hard-code any credentials in our repos. So I would replace the credential with some bash snippet generating a random pw that is also printed to the console for the user of the script to put in their pw manager.

tonsV2: Duly noted! I've updated accordingly :-)

PR #63, opened 2022-07-13
chore: dockerignore scripts, mardown, ...

teleivo:
which is honored by skaffold. So skaffold will not auto-redeploy if we
change for example a .env file in scripts to target a different
environment

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,1 +1,12 @@
-helm/
+.env
+.env.example
+.envrc
+.git
+.github
+.gitignore
+.golangci.yml
+helm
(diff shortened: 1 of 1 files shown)


.dockerignore
```diff
@@ -1 +1,12 @@
+.env
+.env.example
+.envrc
+.git
+.github
+.gitignore
+.golangci.yml
 helm/
```
tonsV2, line 8: For the sake of consistency we should either remove the trailing / here or add it on line 12

tonsV2 (approved):

PR #67, opened 2022-07-16
Fix/missing stack parameters

tonsV2:
I noticed we could unexport StackParameters and included it here but the main focus should be [this](https://github.com/dhis2-sre/im-manager/compare/fix/missing-stack-parameters?expand=1#diff-136f6c444b77ece1ba158828d1f341f38c90c4097dae5da04ed558e3fd695ef8R81).

### DIFF
--- a/pkg/instance/helmfile.go
+++ b/pkg/instance/helmfile.go
@@ -78,7 +78,7 @@ func (h helmfileService) loadStackParameters(folder string, name string) (stackP
 	data, err := os.ReadFile(path)
 	if err != nil {
 		if errors.Is(err, os.ErrNotExist) {
-			return stackParameters{}, err
+			return stackParameters{}, nil
 		}
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


pkg/instance/helmfile.go
```diff
@@ -69,13 +71,17 @@ func (h helmfileService) executeHelmfileCommand(accessToken string, instance *mo
...
-type StackParameters map[string]string
+type stackParameters map[string]string
 
-func (h helmfileService) loadStackParameters(folder string, name string) (StackParameters, error) {
+func (h helmfileService) loadStackParameters(folder string, name string) (stackParameters, error) {
 	environment := h.config.Environment
 	path := fmt.Sprintf("%s/%s/parameters/%s/parameters.yaml", folder, name, environment)
 	data, err := os.ReadFile(path)
 	if err != nil {
+		var e *fs.PathError
```
teleivo, line 81: If we want to return empty stack parameters if the file does not exist then

```go
    if errors.Is(err, os.ErrNotExist) {
        return stackParameters{}, err
    }
```

is more accurate. `os.ReadFile()` will return `permission denied` if the file exists but we do not have permission to read it. In that case `errors.As(err, &e)` will also return true.

teleivo (commented):
If we only want to return empty stack parameters if the file does not exist we should adapt the changes. See code comment 😄

teleivo (approved):
kind of weirdly looking maybe but `return nil, nil`

```go
		if errors.Is(err, os.ErrNotExist) {
			return stackParameters{}, nil
		}
```

would probably work for us 😄

PR #73, opened 2022-07-23
Feat/api cleanup

tonsV2:
* Merge endpoints
* Move parameter encryption into the repository layer


pkg/instance/handler.go
```diff
@@ -102,80 +116,61 @@ func (h Handler) Create(c *gin.Context) {
...
-//   404: Error
-//   415: Error
-func (h Handler) Restart(c *gin.Context) {
-	idParam := c.Param("id")
-	id, err := strconv.ParseUint(idParam, 10, 64)
-	if err != nil {
-		_ = c.AbortWithError(http.StatusBadRequest, fmt.Errorf("failed to parse id: %s", err))
-		return
-	}
+		if sourceInstance.DeployLog == "" {
```
teleivo, line 133: Is that the only information we have to check whether an instance is deployed?

pkg/instance/handler.go
```diff
@@ -102,80 +116,61 @@ func (h Handler) Create(c *gin.Context) {
...
+			err := fmt.Errorf("source instance (%s) not deployed", sourceInstance.Name)
```
teleivo, line 134: For strings that are user provided I would default to using `%q` just in case there are any whitespace issues. You could argue that since we validate using `dns_rfc1035_label` that should not be a problem. I would still default to it as it does not hurt here and reduces the chance of us missing it elsewhere where it hurts.

pkg/instance/handler.go
```diff
@@ -63,8 +66,17 @@ type CreateInstanceRequest struct {
 //   403: Error
 //   404: Error
 //   415: Error
-func (h Handler) Create(c *gin.Context) {
-	var request CreateInstanceRequest
+func (h Handler) Launch(c *gin.Context) {
+	deploy := true
+	if deployParam, ok := c.GetQuery("deploy"); ok {
```
teleivo, line 71: I would prefer having `deploy` in the JSON body as well instead of mixing body/query params.

pkg/instance/repository.go
```diff
@@ -12,22 +14,34 @@ import (
...
 	Unlink(instance *model.Instance) error
-	Create(instance *model.Instance) error
 	Save(instance *model.Instance) error
 	FindWithParametersById(id uint) (*model.Instance, error)
 	FindByNameAndGroup(instance string, group string) (*model.Instance, error)
 	SaveDeployLog(instance *model.Instance, log string) error
 	FindById(id uint) (*model.Instance, error)
 	Delete(id uint) error
 	FindByGroupNames(names []string) ([]*model.Instance, error)
+	FindWithDecryptedParametersById(id uint) (*model.Instance, error)
```
teleivo, line 24: maybe name it `FindByIdWithDecryptedParameters` instead and group it with the `FindById`

pkg/instance/repository.go
```diff
@@ -37,14 +51,11 @@ func (r repository) Link(source *model.Instance, destination *model.Instance) er
...
 	err := r.db.Create(&link).Error
-	if err != nil {
-		var perr *pgconn.PgError
-		if errors.As(err, &perr) && perr.Code == "23505" {
-			return fmt.Errorf("instance (%d) already linked with a stack of type \"%s\"", source.ID, destination.StackName)
-		}
-		return err
+	var perr *pgconn.PgError
+	if errors.As(err, &perr) && perr.Code == "23505" {
+		return fmt.Errorf("instance (%d) already linked with a stack of type \"%s\"", source.ID, destination.StackName)
```
teleivo, line 56: please use `%q` instead of quoting yourself 😄

pkg/instance/repository.go
```diff
@@ -63,20 +74,23 @@ func (r repository) Unlink(instance *model.Instance) error {
...
-
-func (r repository) Create(instance *model.Instance) error {
-	return r.db.Create(&instance).Error
+	return err
 }
 
 func (r repository) Save(instance *model.Instance) error {
+	enrichParameters(instance)
+
+	// TODO: Handle error?
```
teleivo, line 86: should we not at least return the error?

pkg/instance/repository.go
```diff
@@ -128,3 +142,62 @@ func (r repository) FindByGroupNames(names []string) ([]*model.Instance, error)
...
+		instance.OptionalParameters[i].Value = value
+	}
+
+	return nil
+}
+
+// TODO: Rename PopulateRelations? Or something else?
+func enrichParameters(instance *model.Instance) {
+	requiredParameters := instance.RequiredParameters
+	if len(requiredParameters) > 0 {
```
teleivo, line 189: I think 😋 ranging over `instance.RequiredParameters` should not require `len` check https://groups.google.com/g/golang-nuts/c/gVgVLQU1FFE

try https://go.dev/play/p/XcbTTnfzLwM

pkg/instance/handler.go
```diff
@@ -43,16 +43,19 @@ type userClientHandler interface {
...
 
-type CreateInstanceRequest struct {
-	Name      string `json:"name" binding:"required,dns_rfc1035_label"`
-	GroupName string `json:"groupName" binding:"required"`
-	StackName string `json:"stackName" binding:"required"`
+type LaunchInstanceRequest struct {
+	Name               string                            `json:"name" binding:"required,dns_rfc1035_label"`
+	Group              string                            `json:"groupName" binding:"required"`
+	Stack              string                            `json:"stackName" binding:"required"`
+	RequiredParameters []model.InstanceRequiredParameter `json:"requiredParameters"`
```
teleivo, line 50: I would not use our model in the user facing API. I mean in `enrichParameters` you do override `InstanceID`, `StackName` but this opens up a spot that can either introduce security issues or bugs.

pkg/instance/repository.go
```diff
@@ -63,20 +74,23 @@ func (r repository) Unlink(instance *model.Instance) error {
...
+	return err
 }
 
 func (r repository) Save(instance *model.Instance) error {
+	enrichParameters(instance)
+
+	// TODO: Handle error?
+	_ = r.decryptParameters(instance)
+
+	err := r.encryptParameters(instance)
```
teleivo, line 89: why do we decrypt and encrypt them again right away?

pkg/instance/repository.go
```diff
@@ -12,22 +14,34 @@ import (
...
 	FindById(id uint) (*model.Instance, error)
 	Delete(id uint) error
 	FindByGroupNames(names []string) ([]*model.Instance, error)
+	FindWithDecryptedParametersById(id uint) (*model.Instance, error)
 }
 
 type repository struct {
-	db *gorm.DB
+	db     *gorm.DB
+	config config.Config
```
teleivo, line 29: I would only depend on what is needed. This repo does not need the entire config. AFAIK it only needs the `InstanceParameterEncryptionKey` 😄

pkg/instance/handler.go
```diff
@@ -102,80 +116,61 @@ func (h Handler) Create(c *gin.Context) {
...
-		return
+		err = h.instanceService.Link(sourceInstance, savedInstance)
+		if err != nil {
+			_ = c.Error(err)
+			return
+		}
 	}
 
-	instance, err := h.instanceService.FindById(uint(id))
+	savedInstance, err := h.instanceService.Save(instance)
```
teleivo, line 158: not an easy to read diff 😅 so maybe I am wrong, but it looks like we save twice in case we link (see line 145).

pkg/instance/repository.go
```diff
@@ -128,3 +142,62 @@ func (r repository) FindByGroupNames(names []string) ([]*model.Instance, error)
...
+			return err
+		}
+		instance.OptionalParameters[i].Value = value
+	}
+
+	return nil
+}
+
+// TODO: Rename PopulateRelations? Or something else?
+func enrichParameters(instance *model.Instance) {
```
teleivo, line 187: Please declare functions in call order, which makes it easier to read (top down order) 🤓

pkg/instance/service.go
```diff
@@ -74,64 +75,54 @@ type service struct {
 	kubernetesService  kubernetesExecutor
 }
 
-func (s service) LinkDeploy(token string, sourceInstance, destinationInstance *model.Instance) error {
-	err := s.link(sourceInstance, destinationInstance)
+func (s service) ConsumeParameters(source, destination *model.Instance) error {
+	sourceStack, err := s.stackService.Find(source.StackName)
 	if err != nil {
 		return err
```
teleivo, line 81: Since we are using `Find` here for source and destination stack I think we should provide this context to the error we return. Like

`fmt.Errorf("error finding stack %q of source instance: %w", source.StackName, err)`

I am using `%w` here, since we might return our not found error or similar later on. So we still want to check that the root is such an error.

pkg/instance/service.go
```diff
@@ -230,34 +221,23 @@ func (s service) unlink(id uint) error {
...
 
-func (s service) Create(instance *model.Instance) (*model.Instance, error) {
-	err := s.instanceRepository.Create(instance)
-	if err != nil {
-		return nil, err
-	}
-
-	instanceWithParameters, err := s.instanceRepository.FindWithParametersById(instance.ID)
+func (s service) Save(instance *model.Instance) (*model.Instance, error) {
+	enrichParameters(instance)
```
teleivo, line 225: If I read correctly you are calling `enrichParameters` in the repo.Save as well 😅  In that case, I don't think its needed.

scripts/dhis2-launch.sh
```diff
@@ -0,0 +1,55 @@
...
+      \"value\": \"$PGADMIN_INSTALL\"
+    }
+  ],
+  \"requiredParameters\": [
+    {
+      \"name\": \"DATABASE_ID\",
+      \"value\": \"$DATABASE_ID\"
+    }
+  ]
+}" | $HTTP post "$INSTANCE_HOST/instances" "Authorization: Bearer $ACCESS_TOKEN"
```
teleivo, line 55: Since the script is called launch I see `launch` leaking into the language we use with users. I would love for us to reduce the vocabulary to a necessary minimum while obviously still making sense 😋

teleivo (changes requested):
I think we should talk about exposing our internal model to the API and naming before merging the changes 😄

pkg/instance/crypt.go
```diff
@@ -11,7 +11,7 @@ import (
 
 var iv = []byte{83, 108, 97, 118, 97, 32, 85, 107, 114, 97, 105, 110, 105, 33, 33, 33}
 
-func encryptText(text string, key string) (string, error) {
+func encryptText(key string, text string) (string, error) {
```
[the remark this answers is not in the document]
tonsV2: Decryption would fail but not encryption.

I changed from value/key to key/value because I found that more logic. But what do you suggest making this more safe?

pkg/instance/handler.go
```diff
@@ -43,16 +43,19 @@ type userClientHandler interface {
...
-	Name      string `json:"name" binding:"required,dns_rfc1035_label"`
-	GroupName string `json:"groupName" binding:"required"`
-	StackName string `json:"stackName" binding:"required"`
+type LaunchInstanceRequest struct {
+	Name               string                            `json:"name" binding:"required,dns_rfc1035_label"`
+	Group              string                            `json:"groupName" binding:"required"`
+	Stack              string                            `json:"stackName" binding:"required"`
+	RequiredParameters []model.InstanceRequiredParameter `json:"requiredParameters"`
+	OptionalParameters []model.InstanceOptionalParameter `json:"optionalParameters"`
+	SourceInstance     uint                              `json:"sourceInstance"`
```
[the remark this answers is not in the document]
tonsV2: We don't use name/group for instances. We still use its id. Using name/group for instances would require changes to our API gateway which are not trivial.

What kind of alternative name or implementation would you suggest?

pkg/instance/handler.go
tonsV2: Yes, if there's nothing in the log it's not deployed. However this was the first place I had to do this.

We could wrap the evaluation (sourceInstance.DeployLog == "") in a method if you think that's cleaner?

pkg/instance/handler.go
tonsV2: That's where I had it initially but I felt it didn't really belong there. It's metadata compared to the other properties but let's have a chat

pkg/instance/handler.go
tonsV2: Correct! We need to save before "h.instanceService.Link(sourceInstance, savedInstance)" otherwise our FK relationships won't be satisfied.

pkg/instance/crypt.go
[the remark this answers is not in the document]
teleivo: @tonsV2 read the provided link 😋

pkg/instance/handler.go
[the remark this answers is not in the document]
teleivo: | We don't use name/group for instances. We still use its id. Using name/group for instances would require changes to our API gateway which are not trivial.

understood 😄 so lets think about that at a later point 👌🏻 

| What kind of alternative name or implementation would you suggest?

Maybe `linkedInstance`, `linkInstanceSource`, `linkSourceInstance`

pkg/instance/handler.go
teleivo: That's fine. I thought we might have a timestamp.

pkg/instance/handler.go
teleivo: I wonder if we could leverage https://gorm.io/docs/create.html#Create-With-Associations

pkg/instance/crypt.go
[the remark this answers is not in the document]
tonsV2: You want to implement the helper type solution?

pkg/instance/handler.go
teleivo: Investigate this after merging the PR is fine for me 😄

pkg/instance/repository.go
teleivo: we will review this in a separate task

pkg/instance/repository.go
teleivo: you explained it to me 😄

teleivo (approved):
nice 😄

PR #76, opened 2022-07-27
chore: parse stacks template/yaml using Golang parsers DEVOPS-182

teleivo:
I'll walk through the code in our chat 😄


teleivo:
Closing this for now. I can reopen it if we want to use this approach 😄

PR #77, opened 2022-08-02
refactor: update user scripts

tonsV2:
**This PR attempts to include the following updates**
* All scripts are now invoked using the format: ./script-name.sh group name - Group always comes before name which should make the scripts more command line history friendly
* The scripts have been updated to use the new "deploy" endpoint which both create and deploy an instance
* Variable names have been stripped of the "_NAME" postfix. Please consider if there's other ways improve our namving :pray:
* A small script called `stress.sh` has been implemented. The idea is that it can be a starting point for launching multiple instances. And obviously edited per need.
* The `destroy.sh` has been updated so it now can take a variable number of instance names as last argument and therefore works nicely together with the stress script. Please see the readme file for an example of deleting multiple instances.
* Various scripts has been deleted. Should more be deleted? Have too many been deleted?

Please see the readme for various examples and consider adding the following alias to your shell: alias jwt='jq -R \'split(".") | .[1] | @base64d | fromjson\''


radnov (approved):
Looks good to me with the latest changes 👍

PR #86, opened 2022-08-15
chore: scripts should deploy latest default dhis2 image

teleivo:
https://github.com/dhis2/dhis2-core/blob/1ed13727435f11482202a45367f0bb7d566f7c8f/docker/shared/containers-list.sh\#L26-L30
shows tomcat-9.0-jdk11-openjdk-slim is the default and 2.38.1.1 is the latest release


tonsV2 (approved):

tonsV2:
Yeah, reversing the order sounds good

PR #89, opened 2022-08-24
Feat/openstack seed

tonsV2:


stacks/dhis2-db/seed.sh
```diff
@@ -23,11 +23,12 @@ if [ -n "$DATABASE_ID" ]; then
...
   echo "DATABASE_MANAGER_ABSOLUTE_URL: $DATABASE_MANAGER_ABSOLUTE_URL"
 
-  cd /tmp
+  tmp_file=$(mktemp)
+  curl --fail -L "$DATABASE_MANAGER_ABSOLUTE_URL" -H "Authorization: $IM_ACCESS_TOKEN" > "$tmp_file"
 # Try pg_restore... Or gzipped sql
-  (curl --fail -L "$DATABASE_MANAGER_ABSOLUTE_URL" -H "Authorization: $IM_ACCESS_TOKEN" | pg_restore -U postgres -d "$DATABASE_NAME") || \
-  (curl --fail -L "$DATABASE_MANAGER_ABSOLUTE_URL" -H "Authorization: $IM_ACCESS_TOKEN" | gunzip -v -c | psql -U postgres -d "$DATABASE_NAME")
-  cd -
+  (pg_restore -U postgres -d "$DATABASE_NAME" -j 8 "$tmp_file") || \
```
radnov, line 29: Could the number of jobs (`-j 8`) be a problem on servers that only have 2-4 threads (like the nodes we have in the EKS cluster)?

From https://www.postgresql.org/docs/current/app-pgrestore.html
> A good place to start is the number of CPU cores on the server, but values larger than that can also lead to faster restore times in many cases. Of course, values that are too high will lead to decreased performance because of thrashing.

I suppose 8 is not "too high" compared to the 2 and 4 threads available on the EKS nodes, but it's probably worth testing.

radnov (commented):
Looks good, just left one comment for clarification. Also, should these changes be applied to the seed script in the dhis2 stack?

stacks/dhis2-db/seed.sh
tonsV2: How about we make it 4 and call it quits?

radnov: 4 might be a good compromise for the range of threads we have on the different servers. 👍

radnov (approved):

PR #97, opened 2022-09-17
Feat/auto login

tonsV2:
All user scripts dependent on the environment variable ACCESS_TOKEN being populated with a valid access token. Previously this was something the users had to ensure themselves.

This PR introduces the `auth.sh` which, if sourced from one of the user scripts, will cache, update per need and export the ACCESS_TOKEN environment variable automatically.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 .idea
 im-manager
 .env
+.access_token_cache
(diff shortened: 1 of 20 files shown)


radnov (approved):
Very nice 👍

PR #106, opened 2022-10-01
chore: consistently label all stack resources

teleivo:
at least all the resources we have control over. the pgAdmin helm chart only allows us to label pods and not deployments for example.

* return an error when trying to stream logs using a type selector that does not exist in the instances stack. Example `no pod found using the selector: "im-id=52,im-type=dff"`

### DIFF
--- a/pkg/instance/kubernetes.go
+++ b/pkg/instance/kubernetes.go
@@ -176,9 +176,25 @@ func (ks kubernetesService) getPod(instance *model.Instance, typeSelector string
 	return pods.Items[0], nil
 }
 
-func (ks kubernetesService) restart(instance *model.Instance) error {
-	labelSelector := fmt.Sprintf("im-id=%d", instance.ID)
-	listOptions := metav1.ListOptions{LabelSelector: labelSelector}
+func (ks kubernetesService) restart(instance *model.Instance, typeSelector string) error {
+	labels := &metav1.LabelSelector{
+		MatchLabels: map[string]string{
+			"im-id": fmt.Sprint(instance.ID),
+		},
+	}
+	if typeSelector == "" {
+		labels.MatchLabels["im-default"] = "true"
+	} else {
+		labels.MatchLabels["im-type"] = typeSelector
+	}
+
+	labelSelector, err := metav1.LabelSelectorAsSelector(labels)
+	if err != nil {
+		return err
+	}
+	listOptions := metav1.ListOptions{
+		LabelSelector: labelSelector.String(),
+	}
 
 	deployments := ks.client.AppsV1().Deployments(instance.GroupName)
 	deploymentList, err := deployments.List(context.TODO(), listOptions)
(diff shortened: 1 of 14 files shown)


pkg/instance/kubernetes.go
```diff
@@ -141,32 +141,43 @@ func (ks kubernetesService) getLogs(instance *model.Instance, selector string) (
...
+	labelSelector, err := metav1.LabelSelectorAsSelector(labels)
+	if err != nil {
+		return v1.Pod{}, err
+	}
 	listOptions := metav1.ListOptions{
-		LabelSelector: labelSelector,
+		LabelSelector: labelSelector.String(),
 	}
 
 	podList, err := ks.client.CoreV1().Pods("").List(context.TODO(), listOptions)
```
tonsV2, line 164: s/podList/pods/ ? I'm well aware the variable was already there ;-)

pkg/instance/kubernetes.go
```diff
@@ -214,9 +225,6 @@ func (ks kubernetesService) pause(instance *model.Instance) error {
 		}
 	}
 
-	labelSelector = fmt.Sprintf("app.kubernetes.io/instance=%s-database", instance.Name)
-	listOptions = metav1.ListOptions{LabelSelector: labelSelector}
-
 	sets := ks.client.AppsV1().StatefulSets(instance.GroupName)
 	setsList, err := sets.List(context.TODO(), listOptions)
```
tonsV2, line 229: Could we, and should we, adopt the same naming convention as above to get rid of the list suffix?

stacks/dhis2-db/helmfile.yaml
```diff
@@ -7,11 +7,15 @@ releases:
...
     values:
+      - commonLabels:
+          im: "true"
+          im-default: "true"
+          im-id: "{{ requiredEnv "INSTANCE_ID" }}"
+          im-type: "db"
+          im-ttl: "{{ env "INSTANCE_TTL" | default "" }}"
+
       - primary:
           podLabels:
```
tonsV2, line 18: This key no longer have a value, shouldn't we remove it completely?

stacks/dhis2/helmfile.yaml
```diff
@@ -53,10 +55,11 @@ releases:
     version: 11.0.4
     verify: false
     values:
-      - primary:
-          podLabels:
-            im-data-id: "{{ requiredEnv "INSTANCE_ID" }}"
+      - commonLabels:
+          im-id: "{{ requiredEnv "INSTANCE_ID" }}"
+          im-type: "db"
```
tonsV2, line 60: On one hand, I want to choose something less generic such as postgresql. But on the other, I want to cater to less technical users which just knows it's a database. What do you think?

We might also think about how the user even know which types they can query by.

teleivo: I chose db as it's consice. We can go either way

pkg/instance/kubernetes.go
teleivo: sure

pkg/instance/kubernetes.go
teleivo: I don't think this works here as we need the `sets` var to pass it to the scale func
It works above as we just chain the list operation directly on the `Pods()` operation

tonsV2: :+1:

tonsV2:
If I create a database instance using the command `./deploy-dhis2-db.sh whoami db` and then try to link a core instance against it by doing `export MYID=core && ./deploy-dhis2-core.sh whoami db $MYID; read && ./destroy.sh whoami $MYID` it fails to connect to the database because the database name actually is `db-database-postgresql`. This can be fixed by updating the hostname pattern definition in the stack file.

(Be aware that the stack is only parsed if it doesn't exists)

tonsV2 (approved):

PR #121, opened 2022-11-02
Feat/preset

tonsV2:
**Use case:** As a user I'd like to deploy an instance based on a set of predefined parameters.

**Description**
A preset is just an instance which isn't deployed.

You can create one just as you would normally deploy an instance but also pass the query parameter named "preset" with its value st to true.

You can use a preset by specifying the id of the preset during a regular deploy.


radnov (commented):
Looks good to me 👍  There's no mention of the `instances?preset=<bool>` [query parameter](https://swagger.io/docs/specification/describing-parameters/#query-parameters) in the API docs, though. 🤔 

See how the docs look here -  https://api.im.radnov.test.c.dhis2.org/instances/docs#operation/deployInstance

tonsV2:
> Looks good to me +1 There's no mention of the `instances?preset=<bool>` [query parameter](https://swagger.io/docs/specification/describing-parameters/#query-parameters) in the API docs, though. thinking
> 
> See how the docs look here - https://api.im.radnov.test.c.dhis2.org/instances/docs#operation/deployInstance

Good catch! I've updated the docs.go file to include the query param

radnov (approved):

pkg/instance/handler.go
```diff
@@ -167,6 +156,44 @@ func (h Handler) Deploy(c *gin.Context) {
 	c.JSON(http.StatusAccepted, savedInstance)
 }
 
+func (h Handler) consumeParameters(user *models.User, sourceInstanceId uint, instance *model.Instance, preset bool) error {
+	sourceInstance, err := h.instanceService.FindByIdDecrypted(sourceInstanceId)
+	if err != nil {
+		return err
+	}
+
+	if sourceInstance.Preset != preset {
```
teleivo, line 165: Only to make it easier to read I would write this as

`if preset && sourceInstance.Preset != false`

the next one

`if preset && sourceInstance.StackName != instance.StackName`

if its only used for validating the preset request (which judging the error message it is).

teleivo (commented):
Nice idea to reuse (paused) instances for presets 😄 

Regarding adding the boolean flag preset/presets to many methods. I think looking at the changes needed in the methods to implement presets its a valid solution. As soon as we would need another flag we should maybe think about how we can implement different behaviors based on differences in types 🤔 

The only thing I would want to clarify before I give a ✅ 😋 is the query parameter `preset` in relation to `deploy`.

pkg/instance/handler.go
```diff
@@ -76,6 +77,15 @@ func (h Handler) Deploy(c *gin.Context) {
 		}
 	}
 
+	preset := false
+	if presetParam, ok := c.GetQuery("preset"); ok {
```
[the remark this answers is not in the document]
tonsV2: I'll return an error if both are set to true

pkg/instance/handler.go
```diff
@@ -76,6 +77,20 @@ func (h Handler) Deploy(c *gin.Context) {
...
+		var err error
+		if preset, err = strconv.ParseBool(presetParam); err != nil {
+			_ = c.Error(err)
+			return
+		}
+	}
+
+	if deploy && preset {
```
teleivo, line 89: `deploy` is set to `true` by default so this will if I read it correctly return an error even if the user did not set the query parameter `deploy=true`

The error should only be returned if query params are both `deploy=true` and `preset=true`.

The variables `deploy` and `preset` should never both be true. This makes the code further down the line easier IMHO.

pkg/instance/handler.go
```diff
@@ -156,7 +164,13 @@ func (h Handler) Deploy(c *gin.Context) {
 	}
 
 	if deploy {
-		err = h.instanceService.Deploy(token, savedInstance)
+		decryptedInstance, err := h.instanceService.FindByIdDecrypted(i.ID)
```
teleivo, line 167: It looks a bit repetitive having to do the saving, findBy multiple times in this handler.

Did we not already do that above in line 138 https://github.com/dhis2-sre/im-manager/pull/121/files#diff-0a77e42443c6231c5241d66235af8a4e3b08dfe25a653f17991b4b02922df648R138 🤔 

Also https://github.com/dhis2-sre/im-manager/blob/20f3ff12f5d743b1015b04dfa552977ca8adde7b/pkg/instance/service.go#L182-L188 is doing a `s.instanceRepository.Save(instance)`

I wonder if we can somehow reduce the number of calls we make to the DB.

pkg/instance/handler.go
```diff
@@ -167,6 +181,39 @@ func (h Handler) Deploy(c *gin.Context) {
...
+	if err != nil {
+		return err
+	}
+
+	if preset && !sourceInstance.Preset {
+		return apperror.NewUnauthorized(fmt.Sprintf("instance (id: %d) isn't a preset", sourceInstance.ID))
+	}
+
+	if preset && sourceInstance.StackName != instance.StackName {
+		return apperror.NewUnauthorized(fmt.Sprintf("preset stack (%s) doesn't match instance stack (%s)", sourceInstance.StackName, instance.StackName))
```
teleivo, line 195: Maybe we should default to `%q` for printing strings. We actually use it regularly. Maybe not always. That saves you from doing your own quoting or highlighting via `()`.

pkg/instance/handler.go
teleivo: I just saw the changes to the `instanceService` and that you removed these lines. Can you please explain why you removed them?

teleivo (commented):
Thank you for the changes 😄 

I have a new question now that you added changes to the instanceService 😋

pkg/instance/handler.go
tonsV2: Good catch!

teleivo (approved):
The preset related changes LGTM 😄 We can discuss the changes to the instanceService and whether we can save some calls to the DB later on.

PR #134, opened 2022-11-24
feat: Add Liveness probe timeoutSeconds as opt param to stacks and scripts

radnov:


teleivo (approved):

teleivo:
@radnov tests are failing https://github.com/dhis2-sre/im-manager/actions/runs/3539280405/jobs/5940998060#step:6:29

not sure if its because of your change or some other changes to parameters

radnov:
> @radnov tests are failing https://github.com/dhis2-sre/im-manager/actions/runs/3539280405/jobs/5940998060#step:6:29
> 
> not sure if its because of your change or some other changes to parameters

@teleivo The `stack_test.sh` is failing, because the new optional stack parameter `LIVENESS_PROBE_TIMEOUT_SECONDS` is not loaded for the `dhis2` stack in the im-manager DB yet, but it's being used in the `deploy-dhis2.sh` script. 🤔  

This seems like a two-fold problem - on one hand, when we change a stack's parameters we are running the test with the updated changes before they are actually deployed to the targeted `dev` im-manager env. On the other hand - there's no "stack reloading" in case there's new parameters (currently that's done by hand 😅).

PR #136, opened 2022-12-08
feat: Add resource requests env vars for core and DB [DEVOPS-214]

radnov:


Philip-Larsen-Donnelly:
I wonder if we could use different name than DEVELOPMENT_INSTANCE. 
I'm not sure "development" instance necessarily implies that it should be unstable. In our environment most instances are either "development" or "demo" or "ci" instances. I would rather we have a name that implies, unambiguously, that it may be less stable.

Philip-Larsen-Donnelly (approved):

scripts/instances/deploy-dhis2-core.sh
```diff
@@ -22,6 +22,8 @@ CHART_VERSION=${CHART_VERSION:-0.12.1}
 INSTANCE_TTL=${INSTANCE_TTL:-""}
 FLYWAY_MIGRATE_OUT_OF_ORDER=${FLYWAY_MIGRATE_OUT_OF_ORDER:-false}
 FLYWAY_REPAIR_BEFORE_MIGRATION=${FLYWAY_REPAIR_BEFORE_MIGRATION:-false}
+CORE_RESOURCES_REQUESTS_CPU=${CORE_RESOURCES_REQUESTS_CPU:-250m}
```
tonsV2, line 25: Since we're already in the context of core maybe strip the core part of the name

scripts/instances/deploy-dhis2-db-preset.sh
```diff
@@ -12,6 +12,8 @@ NAME=$2
 DATABASE_ID=${DATABASE_ID:-1}
 DATABASE_SIZE=${DATABASE_SIZE:-10Gi}
 INSTANCE_TTL=${INSTANCE_TTL:-""}
+DB_RESOURCES_REQUESTS_CPU=${DB_RESOURCES_REQUESTS_CPU:-250m}
```
tonsV2, line 15: Are you using all the scripts where you're adding the variables? Unless you're actually using it, I think we're unnecessary cluttering the scripts

scripts/instances/deploy-dhis2-db.sh
```diff
@@ -12,6 +12,8 @@ NAME=$2
 DATABASE_ID=${DATABASE_ID:-1}
 DATABASE_SIZE=${DATABASE_SIZE:-5Gi}
 INSTANCE_TTL=${INSTANCE_TTL:-""}
+DB_RESOURCES_REQUESTS_CPU=${DB_RESOURCES_REQUESTS_CPU:-250m}
```
tonsV2, line 15: Same with DB as mentioned above, we're already in that context. No need to add the prefix

tonsV2 (approved):

scripts/instances/deploy-dhis2-db-preset.sh
teleivo: I think @radnov is following the current pattern of declaring all vars even if they have defaults. If that is what you are talking about. I am not sure why we are doing that. Maybe just to make it easier to discover all parameters 🤷🏻

teleivo (approved):

scripts/instances/deploy-dhis2-core.sh
radnov: Sure 👍

scripts/instances/deploy-dhis2-db-preset.sh
tonsV2: I wasn't aware of that pattern 😁

tonsV2:
In regards to the "pattern" of declaring all variables. I've been declaring variables per need and I think that's how we should go about it.

I've been playing with the thought that we should not declare any variables in the scripts but rather pull all the variables prefixed with IM_PARAM_* (or such) from the environment but I haven't thought this fully through and it might not make sense at all :man_shrugging:

tonsV2 (approved):

teleivo (approved):

PR #137, opened 2022-12-12
Feat/integrations

tonsV2:
Instances relies on data from various external sources. This data could be the name of a Docker image or tag. It could also be the ID of the database which should be used to see DHIS2.

This PR includes 5 different integrations which can be tested using the below user scripts

* integrations_databaseId.sh
* integrations_presetId.sh
* integrations_repository.sh
* integrations_sourceId.sh
* integrations_tags.sh


pkg/integration/handler.go
```diff
@@ -0,0 +1,340 @@
...
+	if request.Key == "IMAGE_REPOSITORY" {
+		token, err := getDockerHubToken(h.config.DockerHub.Username, h.config.DockerHub.Password)
+		if err != nil {
+			_ = c.Error(err)
+			return
+		}
+
+		payload := request.Payload.(map[string]any)
+		organization := payload["organization"].(string)
+		images, err := getDockerHubImages(token, organization)
```
teleivo, line 59: I would create a docker hub client that is created once in our main using the config.
The client should then be passed to our handler or to our service which can call `getImages` or `getTags` on it.

radnov (approved):
Looks good to me 👍

pkg/integration/docker.go
```diff
@@ -0,0 +1,150 @@
...
+	"encoding/json"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net/http"
+
+	"github.com/dhis2-sre/im-manager/pkg/config"
+)
+
+func NewDockerHubClient(config config.Config) dockerHubClient {
```
teleivo, line 14: These are the only things

```go
username := d.config.DockerHub.Username
password := d.config.DockerHub.Password
```

the `DockerHubClient` needs. So Username and Password should be the only parameters passed in and fields stored.

pkg/integration/docker.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	return images, nil
+}
+
+func (d dockerHubClient) getToken() (string, error) {
+	username := d.config.DockerHub.Username
+	password := d.config.DockerHub.Password
+
+	url := "https://hub.docker.com/v2/users/login"
```
teleivo, line 116: I would inline the `url` and `contentType`. If you would not than I would at least declare them right before they are used.

pkg/integration/handler.go
```diff
@@ -0,0 +1,214 @@
...
+	"io"
+	"net/http"
+
+	"github.com/dhis2-sre/im-manager/pkg/config"
+
+	"github.com/dhis2-sre/im-manager/internal/handler"
+	"github.com/gin-gonic/gin"
+)
+
+func NewHandler(config config.Config, client DockerHubClient) Handler {
```
teleivo, line 15: I would also just pass in the configs this handler needs instead of the config.

pkg/integration/handler.go
```diff
@@ -0,0 +1,214 @@
...
+	return Handler{config, client}
+}
+
+type Handler struct {
+	config          config.Config
+	dockerHubClient DockerHubClient
+}
+
+type DockerHubClient interface {
```
teleivo, line 24: Did you write a unit test? If not I would not bother with declaring an interface. You have only one implementation and no unit test.

pkg/integration/docker.go
```diff
@@ -0,0 +1,150 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	b, err := io.ReadAll(response.Body)
+	if err != nil {
+		return nil, err
+	}
+
+	body := struct {
```
teleivo, line 49: Itmight make the methods more readable if you declare package level (unexported) types for the payloads like `tagsBody` here.

pkg/integration/handler.go
```diff
@@ -0,0 +1,214 @@
...
+func (h Handler) Integrations(c *gin.Context) {
+	var request Request
+	if err := handler.DataBinder(c, &request); err != nil {
+		_ = c.Error(err)
+		return
+	}
+
+	if request.Key == "IMAGE_REPOSITORY" {
+		payload := request.Payload.(map[string]any)
+		organization := payload["organization"].(string)
```
teleivo, line 56: I think this can easily lead to a panic https://go.dev/tour/methods/15
If payload["organization"] is not a string.

A bit of error handling will probably give a nicer response to the user. Even though it will probably also work like this as Gin does recover from the panic.

pkg/integration/handler.go
```diff
@@ -0,0 +1,214 @@
...
+			return
+		}
+
+		c.JSON(http.StatusOK, images)
+		return
+	}
+
+	if request.Key == "IMAGE_TAG" {
+		payload := request.Payload.(map[string]any)
+		organization := payload["organization"].(string)
```
teleivo, line 69: same here

pkg/integration/handler.go
```diff
@@ -0,0 +1,214 @@
...
+		}
+
+		c.JSON(http.StatusOK, presets)
+		return
+	}
+
+	c.Status(http.StatusNotFound)
+}
+
+func getInstances(token string, url string) (map[uint]string, error) {
```
teleivo, line 138: I would move this below declaration of getDatabases so they are declared in the order they are used.

teleivo (commented):
With `Instances` do you mean the [UI](https://github.com/dhis2-sre/im-web-client)?

Feels a bit odd to have these different things (Docker images, databases, presets, ...) requested/returned from one endpoint. But they can be separated out later if needed 👌🏻

pkg/integration/docker.go
tonsV2: You always get me on this one! Thanks :-)

pkg/integration/docker.go
tonsV2: Content type sure, but the url I declared at top so it could easily be changed. Doesn't that make sense?

pkg/integration/handler.go
tonsV2: org, ok := payload["organization"]
		if !ok {
			_ = c.Error(fmt.Errorf("payload doesn't contain \"organization\""))
			return
		}

		organization, ok := org.(string)
		if !ok {
			_ = c.Error(fmt.Errorf("\"organization\" must be a string"))
			return
		}

Would the above be adequate?

pkg/integration/docker.go
```diff
@@ -0,0 +1,150 @@
...
+
+func NewDockerHubClient(config config.Config) dockerHubClient {
+	return dockerHubClient{config: config}
+}
+
+type dockerHubClient struct {
+	config config.Config
+}
+
+func (d dockerHubClient) GetTags(organization string, repository string) ([]string, error) {
```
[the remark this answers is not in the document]
tonsV2: No! :-)

The organization is hard coded in the stacks.

https://github.com/dhis2-sre/im-manager/blob/master/stacks/dhis2/helmfile.yaml#L9

pkg/integration/docker.go
teleivo: It can easily be changed anywhere 😋 but I'll leave the final spot up to you. I declare vars close to where they are used so their scope is minimal. It also makes it easier to read IMHO.

pkg/integration/docker.go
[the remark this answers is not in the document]
teleivo: This might be clearer if we hard code it in here I think.

teleivo (approved):

pkg/integration/docker.go
tonsV2: The response struct is the same for both tags and images so I created a struct called `dockerHubResponse` which is used by both.

pkg/integration/docker_test.go
```diff
@@ -0,0 +1,86 @@
...
+
+func Test_dockerHubClient_GetImages(t *testing.T) {
+	hubClient := NewDockerHubClient("", "")
+	hubClient.client = &clientMockImages{}
+
+	organization := "dhis2"
+	images, err := hubClient.GetImages(organization)
+	require.NoError(t, err)
+
+	assert.True(t, slices.Contains(images, "core"))
```
teleivo, line 25: Testify has a huge list of assertions we can use. So IMHO we should not add another dependency unless its absolutely needed. Testify is already a pretty big library. As a purist 😋 I would only use go test and https://pkg.go.dev/github.com/google/go-cmp/cmp

Since we are using testify I would reach for 
https://pkg.go.dev/github.com/stretchr/testify/assert#ElementsMatch
since I assume you only expect a slice with 2 elements in it and you don't care about order. If you would only want to assert that a couple of elements are in there but not on the entire slice then https://pkg.go.dev/github.com/stretchr/testify/assert#Contains

pkg/integration/docker_test.go
```diff
@@ -0,0 +1,86 @@
...
+
+func Test_dockerHubClient_GetTags(t *testing.T) {
+	hubClient := NewDockerHubClient("", "")
+	hubClient.client = &clientMockTags{}
+
+	organization := "dhis2"
+	images, err := hubClient.GetImages(organization)
+	require.NoError(t, err)
+
+	assert.True(t, slices.Contains(images, "2.39.0"))
```
teleivo, line 52: same here as above

PR #142, opened 2023-01-12
Feat/tests service

tonsV2:


pkg/instance/service_test.go
```diff
@@ -0,0 +1,87 @@
...
+	repository := &mockInstanceRepository{}
+	repository.
+		On("Link", source, destination).
+		Return(nil)
+
+	service := NewService(config.Config{}, repository, nil, nil, nil)
+
+	err := service.Link(source, destination)
+
+	require.NoError(t, err)
```
teleivo, line 27: Here it might be valuable to use `assert.NoError` to also check the expectations on the mock. Whether to use `require` or `assert` always depends on the answer to "does it make sense to check anything after this did not hold true?" So with pre-conditions/arranging test setup I would likely always use `require`. If we were not able to setup what we need for the test case, there is no point in calling the method under test. After calling the method under test I would likely default to `assert`. 😄

PR #143, opened 2023-01-12
Feat/tests handler

tonsV2:


pkg/instance/handler_test.go
```diff
@@ -0,0 +1,141 @@
...
+	handler := NewHandler(nil, service, nil)
+
+	w := httptest.NewRecorder()
+	c, _ := gin.CreateTestContext(w)
+	c.Set("user", &models.User{Groups: groups})
+
+	handler.List(c)
+
+	assert.Empty(t, w.Body.Bytes())
+	assert.NotEmpty(t, c.Errors)
```
teleivo, line 43: should this contain any specific error?

pkg/instance/handler_test.go
```diff
@@ -0,0 +1,141 @@
...
+	c.Set("user", &models.User{Groups: groups})
+
+	handler.List(c)
+
+	assert.Empty(t, c.Errors)
+	assert.Equal(t, http.StatusOK, w.Code)
+	var body []GroupWithInstances
+	err := json.Unmarshal(w.Body.Bytes(), &body)
+	require.NoError(t, err)
+	assert.Equal(t, groupsWithInstances(groups, instances), body)
```
teleivo, line 82: If you are using this function in your assertion you are not testing it. Maybe we can setup a struct or a raw string with what we expect. A little duplication in tests that with regards to setup is not a problem to me. Testing more without assuming to much about the implementation is more important IMHO. This gives you more confidence and freedom in refactoring 😄

teleivo: So you could either do

```go
    assert.Empty(t, c.Errors)
    assert.Equal(t, http.StatusOK, w.Code)
    var actual []GroupWithInstances
    err := json.Unmarshal(w.Body.Bytes(), &actual)
    require.NoError(t, err)
    expected := []GroupWithInstances{
        {
            Name:     "name",
            Hostname: "hostname",
            Instances: []*model.Instance{
                {
                    Model:     gorm.Model{ID: 1},
                    Name:      "some name",
                    GroupName: "name",
                },
            },
        },
    }
    assert.Equal(t, expected, actual)
```

or

```go
assert.Equal(t, `[{"Name":"name","Hostname":"hostname","Instances":[{"ID":1,"CreatedAt":"0001-01-01T00:00:00Z","UpdatedAt":"0001-01-01T00:00:00Z","DeletedAt":null,"UserID":0,"Name":"some name","GroupName":"name","StackName":"","requiredParameters":null,"optionalParameters":null,"DeployLog":"","Preset":false,"PresetID":0}]}]`, w.Body.String())
```

the latter has the benefit of not having to do the unmarshaling. Since the JSON is not pretty printed its not easy to read and in case the test fails the diff will not be as well.

So I would opt for the first suggestion.

I assume we are asserting some JSON body regularly, so we could write a helper so the test reads nicely.

`assertJSON(t, body, expected)`

we probably also need a struct to unmarshal into.

pkg/instance/handler_test.go
```diff
@@ -0,0 +1,142 @@
...
+
+	require.Empty(t, w.Body.Bytes())
+	require.Len(t, c.Errors, 1)
+	require.ErrorContains(t, c.Errors[0].Err, "some error")
+	repository.AssertExpectations(t)
+}
+
+func newContext(w *httptest.ResponseRecorder, group string) *gin.Context {
+	user := &models.User{
+		ID: uint64(1),
```
teleivo, line 78: I would pass this ID in to the function since this ID needs to match the one in the test setup. Right now that is not obvious 🙂

teleivo (approved):
Great 😃 one small thing I think we should adapt

pkg/instance/handler_test.go
tonsV2: No :smile: That's the user id, it doesn't need to match anything in these tests

PR #159, opened 2023-02-10
chore: cleanup any lingering instances

tonsV2:
@radnov and I have discussed the need for some automatic removal of leftover instances in regards to failing stack tests.

Ideally the tests shouldn't fail but we need more experience to adjust the thresholds so they don't.

The downside of this PR is that we might remove the evidence but on the other hand, if the problem persists, the tests could be ran again locally.

### DIFF
--- a/scripts/stack_tests.sh
+++ b/scripts/stack_tests.sh
@@ -16,8 +16,7 @@ function cleanup_handler {
   exit
 }
 
-trap cleanup_handler INT
-trap cleanup_handler EXIT
+trap cleanup_handler INT EXIT
 
 # Whoami
 INSTANCE_NAME="$INSTANCE_PREFIX-whoami-$INSTANCE_POSTFIX"
--- a/scripts/stack_tests.sh
+++ b/scripts/stack_tests.sh
@@ -7,6 +7,18 @@ GROUP=whoami
 INSTANCE_PREFIX=im-e2e
 INSTANCE_POSTFIX=$(tr -dc '[:lower:]' </dev/urandom | head -c 5; echo '')
 
+function cleanup_handler {
+  # shellcheck disable=SC2046
+  ./destroy.sh whoami $(./list.sh | jq -r '.[].Instances[]?.Name' | grep $INSTANCE_PREFIX | grep "$INSTANCE_POSTFIX")
+  # shellcheck disable=SC2046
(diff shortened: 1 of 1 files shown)


scripts/stack_tests.sh
```diff
@@ -7,6 +7,18 @@ GROUP=whoami
...
+  # shellcheck disable=SC2046
+  ./destroy.sh whoami $(./list.sh | jq -r '.[].Instances[]?.Name' | grep $INSTANCE_PREFIX | grep "$INSTANCE_POSTFIX")
+  # shellcheck disable=SC2046
+  ./destroy.sh whoami $(./listPresets.sh | jq -r '.[].Instances[]?.Name' | grep $INSTANCE_PREFIX | grep "$INSTANCE_POSTFIX")
+  trap - EXIT
+  exit
+}
+
+trap cleanup_handler INT
+trap cleanup_handler EXIT
```
radnov, line 20: The two signals can be `trap`ped on a single line.

radnov (commented):
Looks good. Added a small suggestion 👍

scripts/stack_tests.sh
```diff
@@ -7,6 +7,18 @@ GROUP=whoami
 INSTANCE_PREFIX=im-e2e
 INSTANCE_POSTFIX=$(tr -dc '[:lower:]' </dev/urandom | head -c 5; echo '')
 
+function cleanup_handler {
+  # shellcheck disable=SC2046
+  ./destroy.sh whoami $(./list.sh | jq -r '.[].Instances[]?.Name' | grep $INSTANCE_PREFIX | grep "$INSTANCE_POSTFIX")
```
teleivo, line 12: Just to make sure. I have not executed the bash fully in my head 😂 
Is there any chance we accidentally destroy the wrong instances?

tonsV2: Not unless you create an instance with the same pre- and postfix

tonsV2: To elaborate a bit more here, we have a fairly specific prefix. Even if we ignore the prefix and only focus on the postfix the following should be the case. The postfix consists of 5 lower case alphabetic characters, which gives us 26^5 (== 11881376) combinations. It should be reasonably safe.

teleivo (approved):

radnov (approved):

PR #161, opened 2023-02-10
Feature/get all user instances

tonsV2:
Currently when listing instances, only groups, of which the user is a member, are taken into account.

This is a problem for users such as administrators, and group admins, because such user would get an empty list back.


pkg/instance/repository.go
```diff
@@ -111,12 +112,17 @@ func (r repository) Delete(id uint) error {
...
-	var instances []*model.Instance
-
-	err := r.db.
+	query := r.db.
 		Preload("RequiredParameters.StackRequiredParameter").
-		Preload("OptionalParameters.StackOptionalParameter").
-		Where("group_name IN ?", names).
+		Preload("OptionalParameters.StackOptionalParameter")
+
+	isAdmin := slices.Contains(names, "administrators")
```
teleivo, line 119: Is this a group we create automatically and thus control the group name? If so it would be great to create a link between the code creating and the one here checking the group name. So both places use the exact same string 😄

pkg/instance/service.go
```diff
@@ -265,11 +266,83 @@ func (s service) FindByNameAndGroup(instance string, group string) (*model.Insta
...
+func (s service) filterByGroupId(instances []*model.Instance, test func(instance *model.Instance) bool) (ret []*model.Instance) {
+	for _, instance := range instances {
+		if test(instance) {
+			ret = append(ret, instance)
+		}
+	}
+	return
 }
+
+func (s service) removeDuplicates(groups []*models.Group) []*models.Group {
```
teleivo, line 326: I don't think you are using the receiver 🤔 If so I would remove it. The receiver is just like another parameter. If you can remove it, please look up the call chain if you can remove it there too 😄

pkg/instance/service.go
```diff
@@ -265,11 +266,83 @@ func (s service) FindByNameAndGroup(instance string, group string) (*model.Insta
...
+	if len(groups) < 1 {
```
teleivo, line 327: I think this should be `<=1` 🤔 as in a slice of ` there is no duplicate.

pkg/instance/service.go
```diff
@@ -265,11 +266,83 @@ func (s service) FindByNameAndGroup(instance string, group string) (*model.Insta
...
+	}
+	return groupsWithInstances
+}
+
+func (s service) uniqueInstanceGroups(instances []*model.Instance) []*models.Group {
+	groups := make([]*models.Group, len(instances))
+	for i, instance := range instances {
+		groups[i] = &models.Group{Name: instance.GroupName}
+	}
+	return s.removeDuplicates(groups)
```
teleivo, line 314: Do we want to group instances by group name. If that is the case I wonder if we could write an SQL query to do this work for us. 🤔 Not sure if we would need a new gorm model to represent the particular data we get back from the custom query.

If an SQL query is not an option we could explore using a set like `make(map[string]struct{})` and populate it with the group names. This would take care of removing duplicates 😄

pkg/instance/repository.go
tonsV2: Yes, that is a group that we create automatically but that all takes place in the user service and is defined [here](https://github.com/dhis2-sre/im-user/blob/master/pkg/model/group.go#L9). It would actually be nice if it could be generated as part of the client but I'm not sure how and even if that make sense since the client itself is generated from our open api spec.

It is actually also defined [here](https://github.com/dhis2-sre/im-manager/blob/master/internal/handler/authorize.go#L8) but I don't want the repository to depend on something from the (internal) handler package.

Maybe we could just define it in the repository as it's defined in the handler package as noted above?

pkg/instance/service.go
tonsV2: My reason for define these as methods rather than package global functions is because I want their scope to be limited to the `service` struct.

pkg/instance/service.go
tonsV2: Correct! :-)

pkg/instance/repository.go
```diff
@@ -110,19 +113,73 @@ func (r repository) Delete(id uint) error {
...
+	Hostname  string
+	Instances []*model.Instance
+}
+
+func (r repository) FindByGroups(groups []*models.Group, presets bool) ([]GroupWithInstances, error) {
+	groupsByName := make(map[string]*models.Group)
+	for _, group := range groups {
+		groupsByName[group.Name] = group
+	}
+	groupNames := maps.Keys(groupsByName)
```
teleivo, line 129: You could gather the `groupNames` in your loop 😄 Since you are already looping over all groups and this is the only usage of this dependency. `maps.Keys` will loop again.

I'll leave that up to you 👌🏻

pkg/instance/repository.go
```diff
@@ -110,19 +113,73 @@ func (r repository) Delete(id uint) error {
...
 		Find(&instances).Error
-
 	return instances, err
 }
 
+func mapInstances(groupNames []string, result []*model.Instance) map[string][]*model.Instance {
+	instancesMap := make(map[string][]*model.Instance, len(groupNames))
+	for _, instance := range result {
+		groupName := instance.GroupName
+		instancesMap[groupName] = append(instancesMap[groupName], []*model.Instance{instance}...)
```
teleivo, line 162: I am not sure we need to wrap and spread `instance` here in the append 🤔

pkg/instance/repository.go
```diff
@@ -110,19 +113,73 @@ func (r repository) Delete(id uint) error {
...
+		groupsByName[group.Name] = group
+	}
+	groupNames := maps.Keys(groupsByName)
+
+	instances, err := r.findInstances(groupNames, presets)
+	if err != nil {
+		return nil, err
+	}
+
+	instancesMap := mapInstances(groupNames, instances)
```
teleivo, line 136: How does `groupInstancesByGroup` sound for you? To describe what we map to

pkg/instance/repository.go
```diff
@@ -110,19 +113,73 @@ func (r repository) Delete(id uint) error {
...
+		groupName := instance.GroupName
+		instancesMap[groupName] = append(instancesMap[groupName], []*model.Instance{instance}...)
+	}
+	return instancesMap
+}
+
+func groupWithInstances(instancesMap map[string][]*model.Instance, groupMap map[string]*models.Group) []GroupWithInstances {
+	var groupWithInstances []GroupWithInstances
+	for groupName, instances := range instancesMap {
+		if instances == nil {
```
teleivo, line 170: Could we get rid of `nil` elements by adding a `where` clause to our query?

teleivo (commented):
Great! Just some small suggestions 😄

pkg/instance/repository.go
tonsV2: I think that would make it less readable and since it won't affect performance I'll leave it as is. Thanks for the suggestion though :-)

pkg/instance/repository.go
tonsV2: I don't think so. At least I'm not sure how to do so. We could move the if statement into the "group to instance" function but that would be the same.

pkg/instance/service.go
```diff
@@ -265,11 +265,13 @@ func (s service) FindByNameAndGroup(instance string, group string) (*model.Insta
 	return s.instanceRepository.FindByNameAndGroup(instance, group)
 }
 
-func (s service) FindInstances(groups []*models.Group, presets bool) ([]*model.Instance, error) {
-	groupNames := make([]string, len(groups))
-	for i, group := range groups {
-		groupNames[i] = group.Name
+func (s service) FindInstances(user *models.User, presets bool) ([]GroupWithInstances, error) {
+	allGroups := append(user.Groups, user.AdminGroups...)
```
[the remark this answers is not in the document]
tonsV2: True, I'll keep that in my if it's needed elsewhere

teleivo (approved):

PR #170, opened 2023-02-23
feat: restart without downtime

tonsV2:
Previously we were restarting instances by scaling them to 0 and then back to whatever number of replicas they had before. The drawback of this method is that it comes with a significant amount of downtime.

This PR updates the "kubectl.kubernetes.io/restartedAt" annotation with the current time and thereby triggers an update in the same way as it would be done using `kubectl rollout restart`.

### DIFF
--- a/pkg/instance/kubernetes.go
+++ b/pkg/instance/kubernetes.go
@@ -212,10 +212,8 @@ func (ks kubernetesService) restart(instance *model.Instance, typeSelector strin
 		return fmt.Errorf("multiple deployments found using the selector: %q", selector)
 	}
 
-	name := items[0].Name
-
 	data := fmt.Sprintf(`{"spec": {"template": {"metadata": {"annotations": {"kubectl.kubernetes.io/restartedAt": "%s"}}}}}`, time.Now().Format(time.RFC3339))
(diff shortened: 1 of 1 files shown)


radnov (approved):
Nice one! 👍

pkg/instance/kubernetes.go
```diff
@@ -210,17 +213,8 @@ func (ks kubernetesService) restart(instance *model.Instance, typeSelector strin
...
-	if err != nil {
-		return err
-	}
-
-	// Scale up
-	_, err = scale(deployments, name, prevReplicas)
-
+	data := fmt.Sprintf(`{"spec": {"template": {"metadata": {"annotations": {"kubectl.kubernetes.io/restartedAt": "%s"}}}}}`, time.Now().Format(time.RFC3339))
+	_, err = deployments.Patch(context.TODO(), items[0].Name, types.StrategicMergePatchType, []byte(data), metav1.PatchOptions{})
 	return err
```
teleivo, line 218: I think we should return our own error. Maybe

`fmt.Errorf("error restarting %q: %v", items[0].Name, err)`

like you said its sometimes hard to find the source of the error. I think one cause of this is that we miss creating our own errors when getting errors from the stdlib or 3rd party libs.

teleivo (approved):
Great 😄

PR #171, opened 2023-02-24
Feature/security checks

tonsV2:
The goal of this PR is to add security checks by using pre-commit hooks.

Several updates and code changes were needed to make the tests pass and the reasoning is the same as [here](https://github.com/dhis2-sre/im-database-manager/pull/235).

Please confirm that disabling gosec is acceptable [here](https://github.com/dhis2-sre/im-manager/pull/171/files#diff-136f6c444b77ece1ba158828d1f341f38c90c4097dae5da04ed558e3fd695ef8R67) and [here](https://github.com/dhis2-sre/im-manager/pull/171/files#diff-136f6c444b77ece1ba158828d1f341f38c90c4097dae5da04ed558e3fd695ef8R79).

And also confirm that disabling go-critic is acceptable [here](https://github.com/dhis2-sre/im-manager/pull/171/files#diff-61861f18b6590831f09bb1a529528aecc218995e4f896f568dec815cb3828ef2R269). An issue about using append to concatenate slices can be found [here](https://github.com/go-critic/go-critic/issues/865). (Although I didn't find it of much use :-))

### DIFF
--- a/.github/workflows/cicd.yaml
+++ b/.github/workflows/cicd.yaml
@@ -14,7 +14,7 @@ on:
 
 jobs:
   call-workflow:
-    uses: dhis2-sre/gha-workflows/.github/workflows/instance-manager.yaml@v0.5.0
+    uses: dhis2-sre/gha-workflows/.github/workflows/instance-manager.yaml@v0.8.0
     with:
       PROCESS_NAME: im-manager
       POD_NAME: im-manager
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,21 +1,30 @@
+fail_fast: true
+
 default_install_hook_types:
   - pre-commit
   - commit-msg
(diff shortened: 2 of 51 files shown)


teleivo (approved):
Nice 😄 

I would prefer not to use fail_fast: true https://github.com/dhis2-sre/im-inspector/pull/56/files#diff-63a9c44a44acf85fea213a857769990937107cf072831e1a26808cfde9d096b9R1

As with running tests I'd rather see all the failed tests. This makes it easier to fix all issues at once. In my experience fail fast leads to a game of Whac-A-Mole 😁 

Unless your reason to use it convinces me 😋

tonsV2:
My reason merely is that I feel our pre-commit check takes quite a while to run and I'd appreciate failing fast. I'm not sure if that'll convince you :-)

PR #182, opened 2023-03-10
chore: add tests and make the KubernetesService testable

tonsV2:
The KubernetesService required a Kubernetes server being available. This PR changes that by loading the configuration on demand and add some tests.


teleivo:
Should we close this as we've since changed our testing strategy and have integration tests at least for part of the k8s service?

PR #184, opened 2023-03-16
chore: initial merge of the database manager

tonsV2:
# TODO/Considerations
* authorize.go - Maybe rewrite, maybe move into package... Database specific into database package and so on
* Integrations does http request to itself... Not ideal
* Engine... Routes should also be the package which they are setting up
* ~~Global config... Probably not, database config is already in the database package. Same for instance config? Maybe it is better to keep this globally so we're not scattering the configuration all over the app which could make it difficult to actually know how to configure it~~
* ~~User scripts... One big pile or somehow grouped?~~

Beside the user scripts I think the best approach would be to make Jira tasks for the above and deal with them separately of this PR. Let me know what you guys think.


radnov:
> * Global config... Probably not, database config is already in the database package. Same for instance config? Maybe it is better to keep this globally so we're not scattering the configuration all over the app which could make it difficult to actually know how to configure it

Global config does seem like an "easier" option on the surface, but what would be the downsides compared to having config per package? 👀 


> * User scripts... One big pile or somehow grouped?

I think we should at least keep them in separate directories, under `/scripts` 🤔 


> Beside the user scripts I think the best approach would be to make Jira tasks for the above and deal with them separately of this PR. Let me know what you guys think.

Yeah, having these TODOs as tasks in Jira sounds like a good idea to me. Keeping the PRs smaller would be nice, even though merging the services into one does entail larger PRs in general.

tonsV2:
> > * Global config... Probably not, database config is already in the database package. Same for instance config? Maybe it is better to keep this globally so we're not scattering the configuration all over the app which could make it difficult to actually know how to configure it
> 
> Global config does seem like an "easier" option on the surface, but what would be the downsides compared to having config per package? eyes
> 

As far as I can tell, only that you can't see the entire config. I've already merged the database config into the existing one because it was a bad idea.

> > * User scripts... One big pile or somehow grouped?
> 
> I think we should at least keep them in separate directories, under `/scripts` thinking
> 

The user scripts are now under /scripts/databases and /scripts/instances.

> > Beside the user scripts I think the best approach would be to make Jira tasks for the above and deal with them separately of this PR. Let me know what you guys think.
> 
> Yeah, having these TODOs as tasks in Jira sounds like a good idea to me. Keeping the PRs smaller would be nice, even though merging the services into one does entail larger PRs in general.

:+1:

pkg/instance/handler.go
```diff
@@ -331,7 +331,7 @@ func (h Handler) Pause(c *gin.Context) {
 	//	404: Error
 	//	415: Error
 	idParam := c.Param("id")
-	id, err := strconv.ParseUint(idParam, 10, 64)
+	id, err := strconv.ParseUint(idParam, 10, 32)
```
teleivo, line 334: why is that necessary 🤔

teleivo (approved):
I would vote for merging if the functionality works. We can rearrange the code later on 😄

PR #193, opened 2023-03-22
chore: define database related routes in the database package

tonsV2:


teleivo (approved):
Makes sense to me 😄

tonsV2:
Great! I'll apply the same pattern for the rest of the routes

PR #242, opened 2023-05-11
refactor: match instance parameters to stack parameters before saving

radnov:


pkg/instance/service.go
```diff
@@ -184,8 +188,55 @@ func (s service) unlink(id uint) error {
 	return s.instanceRepository.Unlink(instance)
 }
 
+func (s service) matchRequiredParameters(stackParams []model.StackRequiredParameter, instanceParams []model.InstanceRequiredParameter) error {
+	stackRequiredParameterNames := make([]string, len(stackParams))
```
teleivo, line 192: Names in Go are usually chosen to be shorter than what I am used to from other languages. See https://github.com/golang/go/wiki/CodeReviewComments#variable-names (https://go.dev/doc/effective_go#names)

Maybe

```suggestion
	paramNames := make([]string, len(stackParams))
```

would do as well 😄

pkg/instance/service.go
```diff
@@ -184,8 +188,55 @@ func (s service) unlink(id uint) error {
...
+
+	for i, param := range stackParams {
+		stackRequiredParameterNames[i] = param.Name
+	}
+
+	for _, param := range instanceParams {
+		if !slices.Contains(stackRequiredParameterNames, param.StackRequiredParameterID) {
```
teleivo, line 199: I would go for a set of names which allows for easy and fast lookup. Right now we are looping over the params for every param (O(N^2)). It might not matter for the amount of parameters we have though.

teleivo (commented):
Nice 😄 

I would present all errors to the user in a validation. Right now they will have to play HTTP Whac-A-Mole in case there are multiple unknown parameters as we return on the first error.

pkg/instance/service.go
teleivo: `paramNames := make(map[string]struct{})` is the way (using less memory than `bool`)


`paramNames[param.Name] = struct{}{}` (https://www.scaler.com/topics/golang/golang-set/)

pkg/instance/service.go
```diff
@@ -184,8 +186,60 @@ func (s service) unlink(id uint) error {
 	return s.instanceRepository.Unlink(instance)
 }
 
+func (s service) matchRequiredParameters(stackParams []model.StackRequiredParameter, instanceParams []model.InstanceRequiredParameter) []string {
```
teleivo, line 189: `(s service)` is like another parameter to `matchRequiredParameters`. So we treat it like any unused parameter and remove it i.e. make it a function instead of method 😄

pkg/instance/service.go
```diff
@@ -184,8 +186,60 @@ func (s service) unlink(id uint) error {
...
+	unmatchedReqParams := s.matchRequiredParameters(instanceStack.RequiredParameters, instance.RequiredParameters)
+	unmatchedOptParams := s.matchOptionalParameters(instanceStack.OptionalParameters, instance.OptionalParameters)
+
+	unmatchedParams := make([]string, 0)
+	unmatchedParams = append(unmatchedParams, unmatchedReqParams...)
+	unmatchedParams = append(unmatchedParams, unmatchedOptParams...)
+
+	if len(unmatchedParams) > 0 {
+		return nil, apperror.NewBadRequest(
+			fmt.Sprintf("these parameters were not found for the \"%s\" stack: %v", instance.StackName, unmatchedParams),
```
teleivo, line 238:
```suggestion
			fmt.Sprintf("parameters %q are not valid parameters for stack %q", instance.StackName, unmatchedParams),
```

`%q` quotes for us 😄 https://pkg.go.dev/fmt#hdr-Printing

pkg/instance/service.go
```diff
@@ -184,8 +186,60 @@ func (s service) unlink(id uint) error {
...
+}
+
 func (s service) Save(instance *model.Instance) (*model.Instance, error) {
-	err := s.instanceRepository.Save(instance)
+	instanceStack, err := s.stackService.Find(instance.StackName)
+	if err != nil {
+		return nil, err
+	}
+
+	unmatchedReqParams := s.matchRequiredParameters(instanceStack.RequiredParameters, instance.RequiredParameters)
```
teleivo, line 229: To make `save` more readable you can put all the unmatched params logic into one function so this here reads like

```go

err:= validateParameters(instanceStack, instance)
if err != nil {
  return err
}
```

teleivo (approved):

tonsV2 (approved):
The rest of the code base is using parameter and not param as variable name. For the sake of consistency... You know what I'm suggesting :wink:

teleivo:
> The rest of the code base is using parameter and not param as variable name. For the sake of consistency... You know what I'm suggesting wink

change the rest of the codebase 😂

tonsV2:
> > The rest of the code base is using parameter and not param as variable name. For the sake of consistency... You know what I'm suggesting wink
> 
> change the rest of the codebase joy

PR's are accepted :wink:

PR #248, opened 2023-05-24
Feature/save

tonsV2:
This PR implements saving of a database. Unlike the already implemented "save as", save replaces the existing database.

The steps to do so are as follows

* Lock database if not already locked
* Save as
* Delete the existing database
* Copy the saved as database on S3
* Overwrite the saved as database's properties with the ones from the old database (including the id)
* Unlock the database if it wasn't already locked

During the development of this PR I realized that we have a bug in the "save as" method preventing saving in the format of sql.gz. Therefor this PR only works when dealing with the pgc format.


radnov (approved):
Tested locally and I was able to save a database successfully. Approving with some comments/suggestions.

pkg/database/handler.go
```diff
@@ -221,6 +228,114 @@ func (h Handler) SaveAs(c *gin.Context) {
...
+		if err != nil {
+			_ = c.Error(err)
+			return
+		}
+	}
+
+	c.Status(http.StatusAccepted)
+}
+
+func randomString(length int) string {
```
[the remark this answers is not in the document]
tonsV2: Nope, not as far as I know! But it sure would be nice

pkg/database/handler.go
```diff
@@ -221,6 +228,114 @@ func (h Handler) SaveAs(c *gin.Context) {
...
+		return
+	}
+
+	databaseIdString, err := findParameter("DATABASE_ID", instance, stack)
+	if err != nil {
+		_ = c.Error(err)
+		return
+	}
+
+	databaseId, err := strconv.ParseUint(databaseIdString, 10, 32)
```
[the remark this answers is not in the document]
tonsV2: In case the string doesn't contain an unsigned integer or one which is bigger than 32 bits

pkg/database/handler.go
```diff
@@ -221,6 +228,123 @@ func (h Handler) SaveAs(c *gin.Context) {
...
+	// Responses:
+	//	202:
+	//	401: Error
+	//	403: Error
+	//	404: Error
+	//	415: Error
+	instanceIdParam := c.Param("instanceId")
+	instanceId, err := strconv.ParseUint(instanceIdParam, 10, 32)
+	if err != nil {
+		badRequest := apperror.NewBadRequest("error parsing instanceId")
```
teleivo, line 260: Now that the user service merged with the manager we should use the same error handling approach

https://github.com/dhis2-sre/im-manager/blob/31539cd12138b81a4d6243a1ac10a29ac71664fd/pkg/user/handler.go#L232-L233

pkg/database/handler.go
```diff
@@ -221,6 +228,123 @@ func (h Handler) SaveAs(c *gin.Context) {
...
+	}
+
+	databaseId, err := strconv.ParseUint(databaseIdString, 10, 32)
+	if err != nil {
+		badRequest := apperror.NewBadRequest("error parsing databaseId")
+		_ = c.Error(badRequest)
+		return
+	}
+
+	database, err := h.databaseService.FindById(uint(databaseId))
```
teleivo, line 290: You are making several calls to the databaseService which to me suggests we should create a method in the databaseService that takes care of 

* h.databaseService.FindById(uint(databaseId))
* ideally check access
* h.databaseService.Lock(uint(databaseId), uint(instanceId), user.ID)
* h.databaseService.Save(database, instance, stack)
* h.databaseService.Unlock(uint(databaseId))

or move this into `Save`

pkg/database/handler.go
```diff
@@ -221,6 +228,123 @@ func (h Handler) SaveAs(c *gin.Context) {
...
+
+	if !isLocked {
+		_, err := h.databaseService.Lock(uint(databaseId), uint(instanceId), user.ID)
+		if err != nil {
+			_ = c.Error(err)
+			return
+		}
+	}
+
+	err = h.databaseService.Save(database, instance, stack)
```
teleivo, line 324: It looks to me as if we are potentially leaving the DB in a locked state if any of the operations from here on return an error. We should probably unlock in a defer.

pkg/database/handler.go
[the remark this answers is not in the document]
teleivo: Can we use a UUID instead of `randomString`? If not I think `randomString` its only used in the service so should be declared there 😄

pkg/database/service.go
```diff
@@ -266,7 +270,66 @@ func (s service) FindExternalDownload(uuid uuid.UUID) (model.ExternalDownload, e
...
+			return
+		}
+
+		err = s.repository.UpdateId(saved.ID, database.ID)
+		if err != nil {
+			logError(err)
+			return
+		}
+	})
+	if err != nil {
```
teleivo, line 318: please remove as this is already covered by the final return

pkg/database/service.go
```diff
@@ -266,7 +270,66 @@ func (s service) FindExternalDownload(uuid uuid.UUID) (model.ExternalDownload, e
...
 
-func (s service) SaveAs(database *model.Database, instance *model.Instance, stack *model.Stack, newName string, format string) (*model.Database, error) {
+func (s service) Save(database *model.Database, instance *model.Instance, stack *model.Stack) error {
+	tmpName := randomString(20)
+	format := getFormat(database)
+	_, err := s.SaveAs(database, instance, stack, tmpName, format, func(saved *model.Database) {
+		err := s.Delete(database.ID)
+		if err != nil {
+			logError(err)
+			return
```
teleivo, line 280: I think you are discarding the error since you are not returning it `Save` will return nil

pkg/database/service.go
```diff
@@ -266,7 +270,66 @@ func (s service) FindExternalDownload(uuid uuid.UUID) (model.ExternalDownload, e
...
+		}
+
+		u, err := url.Parse(saved.Url)
```
teleivo, line 283: We could move this before `s.Delete` so we don't delete in case we cannot parse the new saved DB 🤔

pkg/database/handler.go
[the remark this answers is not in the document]
radnov: Right, but could we ever end up in this situation? just wondering 🤔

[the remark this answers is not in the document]
tonsV2: Probably not but we should still handle the error

pkg/database/service.go
tonsV2: This handles any error potentially returned from the "save as"

pkg/database/service.go
tonsV2: Yeah, but that's because it's running in a go routine and it's likely to happen after the request is done

pkg/database/service.go
teleivo:
```go
if err != nil {
  return err
}

return err
```

should be

```go
return err
```

pkg/database/service.go
teleivo: So the save (saveAs) request is async. So users don't really know if saving/saving as actually worked right?

pkg/database/service.go
tonsV2: Correct! I get it now, thanks :-)

pkg/database/service.go
tonsV2: No, not until the saving has completed... Or failed. There's a todo comment within the logError function stating that we actually should persist the error message somewhere (possible on the database entity)

pkg/database/handler.go
tonsV2: I've moved the locking handling into the service function. As we have talked about elsewhere, I'm not completely against moving more logic into the service. However I'd like to investigate, discuss it further and not make it part of this PR.

pkg/database/handler.go
```diff
@@ -221,6 +225,99 @@ func (h Handler) SaveAs(c *gin.Context) {
 	c.JSON(http.StatusCreated, save)
 }
 
+// Save database
+func (h Handler) Save(c *gin.Context) {
+	// swagger:route POST /databases/save/{instanceId} saveDatabase
+	//
+	// Save database
+	//
+	// Saving a database involves the following steps
```
teleivo, line 234: I don't think we need this doc 😄  This is clear from the `Save` method in the service. If you want to describe this than it should be on the `Save` method in the service

tonsV2: The benefit of putting it on the handler is that it'll be included in our swagger generated api documentation. But maybe we don't need it after all. One downside is that we have to keep the comment in sync with the code.

@radnov, @Philip-Larsen-Donnelly - What do you guys think?

teleivo: These are implementation details to me. Users of the API should care about the end result not how we achieve it. This is why I would not document the steps. You also duplicate the code you wrote and thus get into the trouble of it getting out of sync.

pkg/database/handler.go
```diff
@@ -221,6 +225,90 @@ func (h Handler) SaveAs(c *gin.Context) {
...
+	// Responses:
+	//	202:
+	//	401: Error
+	//	403: Error
+	//	404: Error
+	//	415: Error
+	instanceIdParam := c.Param("instanceId")
+	instanceId, err := strconv.ParseUint(instanceIdParam, 10, 32)
+	if err != nil {
+		_ = c.AbortWithError(http.StatusBadRequest, errors.New("error parsing instanceId"))
```
teleivo, line 248: I think this (and occurrences below) can be replaced with the GetParam... helper 😄

pkg/database/handler.go
```diff
@@ -221,6 +225,90 @@ func (h Handler) SaveAs(c *gin.Context) {
...
+
+	user, err := handler.GetUserFromContext(c)
+	if err != nil {
+		_ = c.Error(err)
+		return
+	}
+
+	lock := database.Lock
+	isLocked := lock != nil
+	if isLocked && (lock.InstanceID != uint(instanceId) || lock.UserID != user.ID) {
```
teleivo, line 297: This belongs in the `Save` IMHO

pkg/database/handler.go
```diff
@@ -221,6 +225,90 @@ func (h Handler) SaveAs(c *gin.Context) {
...
+
+	databaseIdString, err := findParameter("DATABASE_ID", instance, stack)
+	if err != nil {
+		_ = c.Error(err)
+		return
+	}
+
+	databaseId, err := strconv.ParseUint(databaseIdString, 10, 32)
+	if err != nil {
+		badRequest := apperror.NewBadRequest("error parsing databaseId")
```
teleivo, line 272: do you need to rebase? as you removed `apperror` 🤔

pkg/database/repository.go
```diff
@@ -65,17 +65,17 @@ func (r repository) FindBySlug(slug string) (*model.Database, error) {
 	return d, err
 }
 
-func (r repository) Lock(id, instanceId, userId uint) (*model.Lock, error) {
+func (r repository) Lock(databaseId, instanceId, userId uint) (*model.Lock, error) {
```
teleivo, line 68: Should we stick to `databaseID`? our field names do seem to follow https://github.com/golang/go/wiki/CodeReviewComments#initialisms

tonsV2: I guess we should but it's only a handful of places that we actually do

PR #249, opened 2023-05-24
feat: delete pvc's when destroying an instance

tonsV2:
The title should sum up the purpose of this PR but let me know if there's any questions

### DIFF
--- a/pkg/instance/kubernetes.go
+++ b/pkg/instance/kubernetes.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"log"
 	"os"
 	"os/exec"
 	"strings"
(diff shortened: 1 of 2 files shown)


pkg/instance/kubernetes.go
```diff
@@ -238,6 +239,32 @@ func (ks kubernetesService) resume(instance *model.Instance) error {
...
+	pvcNamePatterns := storageMap[instance.StackName]
+	if pvcNamePatterns == nil {
+		return nil
+	}
+
+	for _, pattern := range pvcNamePatterns {
+		name := fmt.Sprintf(pattern, instance.Name)
+		err := pvcs.Delete(context.TODO(), name, metav1.DeleteOptions{})
+		if err != nil {
+			log.Printf("failed to delete pvc %d: %v", instance.ID, err)
```
teleivo, line 259: Why are you logging the id on error and the name on success?

pkg/instance/kubernetes.go
```diff
@@ -238,6 +239,32 @@ func (ks kubernetesService) resume(instance *model.Instance) error {
 	return nil
 }
 
+func (ks kubernetesService) deletePersistentVolumeClaim(instance *model.Instance) error {
+	// TODO: This should be stack metadata
```
teleivo, line 243: Why is this not part of implementing the feature?

pkg/instance/kubernetes.go
tonsV2: I removed the logging and now just break on first error

PR #250, opened 2023-05-25
Feature/reset

tonsV2:
This PR will allow us to reset an instance.

The reset is performed by destroying the instance in the cluster and redeploying it afterwards.


pkg/instance/handler.go
```diff
@@ -379,6 +380,64 @@ func (h Handler) Pause(c *gin.Context) {
...
+		return
+	}
+
+	user, err := handler.GetUserFromContext(c)
+	if err != nil {
+		_ = c.Error(err)
+		return
+	}
+
+	instance, err := h.instanceService.FindById(uint(id))
```
teleivo, line 419: We call a findBy in reset as well. Can we move the ACL check CanWriteInstance into the service and save one of the findBy calls?

pkg/instance/service.go
```diff
@@ -361,3 +339,54 @@ func (s service) FindInstances(user *model.User, presets bool) ([]GroupWithInsta
...
+	}
+
+	err = s.destroy(token, decrypted)
+	if err != nil {
+		return err
+	}
+
+	err = s.Deploy(token, decrypted)
+	if err != nil {
+		return err
```
teleivo, line 356: What should a user do if deploying again fails? I think we should return that as part of the error 😄 Since we are not retrying to deploy for them.

pkg/instance/handler.go
tonsV2: I'm not sure if that belongs in the service but we can discuss this further. However I think it's out of the scope for this PR.

pkg/instance/service.go
tonsV2: But what should a user do?

radnov (approved):
I have not tested it locally, but I'll approve assuming it does. 👍

PR #251, opened 2023-05-30
chore: use helper function

tonsV2:
Quite a few places we have code like below

	idParam := c.Param("id")
	id, err := strconv.ParseUint(idParam, 10, 32)
	if err != nil {
		_ = c.AbortWithError(http.StatusBadRequest, errors.New("error parsing id"))
		return
	}

Not only is it a bit annoying to repeat but `id` is of the type `uint64` and it has to be converted to `uint` since that's what's most of our methods accepts.

Due to the above I want to introduce a helper function just to make the code a bit more clean.


internal/handler/getPathParameter.go
```diff
@@ -0,0 +1,19 @@
...
+
+import (
+	"fmt"
+	"net/http"
+	"strconv"
+
+	"github.com/gin-gonic/gin"
+)
+
+func GetPathParameter(c *gin.Context, parameter string) (uint, bool) {
```
teleivo, line 11: I would favor using `param/Param` everywhere.

pkg/user/handler.go
```diff
@@ -227,10 +225,8 @@ func (h Handler) FindById(c *gin.Context) {
 	//	403: Error
 	//	404: Error
 	//	415: Error
-	idParam := c.Param("id")
-	id, err := strconv.ParseUint(idParam, 10, 32)
-	if err != nil {
-		_ = c.AbortWithError(http.StatusBadRequest, errors.New("error parsing id"))
+	id, ok := handler.GetPathParameter(c, "id")
```
teleivo, line 228: type conversion in `userWithGroups, err := h.userService.FindById(uint(id))` is not necessary anymore 😄

internal/handler/getPathParameter.go
```diff
@@ -0,0 +1,19 @@
+package handler
```
teleivo, line 1: I would call the file `param.go` so we avoid creating too many little files. `param.go` could be the home for other parameter related helpers.

teleivo (commented):
Great 😄

pkg/user/handler.go
tonsV2: Yes, that's the whole point! :-)

internal/handler/getPathParameter.go
tonsV2: I renamed the file to `pathParameter.go` for now. If we implement more parameter related functions we can make it more generic

internal/handler/getPathParameter.go
tonsV2: I favor consistency and I don't care much either way but feel free to submit a PR

internal/handler/getPathParameter.go
teleivo: To me a more generic name like `param.go` attracts more related code so I would do this from the start.

radnov (approved):

radnov (approved):

teleivo:
Can we first get https://github.com/dhis2-sre/im-manager/pull/259 merged so we don't have to review the same files again 😅

tonsV2:
> Can we first get #259 merged so we don't have to review the same files again sweat_smile

Done!

internal/handler/pathParameter_test.go
```diff
@@ -0,0 +1,28 @@
...
+	id, ok := GetPathParameter(ctx, "id")
+	assert.True(t, ok)
+	assert.Equal(t, uint(123), id)
+}
+
+func TestGetPathParameter_NotFound(t *testing.T) {
+	w := httptest.NewRecorder()
+	ctx, _ := gin.CreateTestContext(w)
+
+	id, ok := GetPathParameter(ctx, "id")
```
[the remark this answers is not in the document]
tonsV2: Why? What does that have to do with this? :smile:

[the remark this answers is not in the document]
teleivo: because you are setting (sending) the status code to 400 in `GetPathParameter`

`_ = c.AbortWithError(http.StatusBadRequest, fmt.Errorf("error parsing %q: %v", parameter, err))`

which is why clients of `GetPathParameter` only return and don't have to set any error like we do in other places.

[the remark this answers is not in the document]
tonsV2: Yes, of course 👍🏻

PR #253, opened 2023-05-31
feat: implement listing of users for admins users

radnov:
This PR adds an admin user endpoint `/users` for listing all users with the groups they belong to.

Example:
```
$ export ACCESS_TOKEN=$(./signInAdmin.sh | jq -r '.access_token') && ./list.sh

[
  {
    "ID": 1,
    "CreatedAt": "2023-05-12T16:05:42.587159Z",
    "UpdatedAt": "2023-05-12T16:05:42.587159Z",
    "DeletedAt": null,
    "Email": "andreas@dhis2.org",
    "Groups": [
      {
        "Name": "administrators",
        "Hostname": "",
        "Users": null,
        "clusterConfiguration": null,
        "CreatedAt": "2023-05-12T16:05:42.588871Z",
        "UpdatedAt": "2023-05-31T10:32:00.126074Z",
        "DeletedAt": null
      }
    ],
    "AdminGroups": []
  },
  {
    "ID": 2,
    "CreatedAt": "2023-05-12T16:09:38.779716Z",
    "UpdatedAt": "2023-05-12T16:09:38.779716Z",
    "DeletedAt": null,
    "Email": "radoslav@dhis2.org",
    "Groups": [
      {
        "Name": "whoami",
        "Hostname": "whoami.im.radnov.test.c.dhis2.org",
        "Users": null,
        "clusterConfiguration": null,
        "CreatedAt": "2023-05-12T16:05:42.594679Z",
        "UpdatedAt": "2023-05-31T10:01:44.49286Z",
        "DeletedAt": null
      }
    ],
    "AdminGroups": [
      {
        "Name": "whoami",
        "Hostname": "whoami.im.radnov.test.c.dhis2.org",
        "Users": null,
        "clusterConfiguration": null,
        "CreatedAt": "2023-05-12T16:05:42.594679Z",
        "UpdatedAt": "2023-05-31T10:01:44.49286Z",
        "DeletedAt": null
      }
    ]
  },
  {
    "ID": 3,
    "CreatedAt": "2023-05-26T16:05:53.368246Z",
    "UpdatedAt": "2023-05-26T16:05:53.368246Z",
    "DeletedAt": null,
    "Email": "test@dhis2.org",
    "Groups": [
      {
        "Name": "whoami",
        "Hostname": "whoami.im.radnov.test.c.dhis2.org",
        "Users": null,
        "clusterConfiguration": null,
        "CreatedAt": "2023-05-12T16:05:42.594679Z",
        "UpdatedAt": "2023-05-31T10:01:44.49286Z",
        "DeletedAt": null
      }
    ],
    "AdminGroups": []
  }
]
```

### DIFF
--- a/pkg/user/handler.go
+++ b/pkg/user/handler.go
@@ -31,6 +31,7 @@ type userService interface {
 	SignUp(email string, password string) (*model.User, error)
 	SignIn(email string, password string) (*model.User, error)
 	FindById(id uint) (*model.User, error)
+	FindAll() ([]*model.User, error)
 }
 
 type tokenService interface {
@@ -237,4 +238,30 @@ func (h Handler) FindById(c *gin.Context) {
 	}
 
 	c.JSON(http.StatusOK, userWithGroups)
+}
+
+// FindAll user
+func (h Handler) FindAll(c *gin.Context) {
+	// swagger:route GET /users findAllUsers
+	//
+	// Find users
+	//
+	// Find all users with the groups they belong to
+	//
+	// security:
+	//	oauth2:
+	//
+	// responses:
+	//	200: []User
+	//	401: Error
+	//	403: Error
+	//	404: Error
+	//	415: Error
+	users, err := h.userService.FindAll()
+	if err != nil {
+		_ = c.Error(err)
+		return
+	}
+
+	c.JSON(http.StatusOK, users)
 }
--- a/pkg/user/repository.go
+++ b/pkg/user/repository.go
@@ -2,6 +2,7 @@ package user
 
 import (
 	"errors"
+	"fmt"
 
 	"github.com/dhis2-sre/im-manager/internal/errdef"
 
@@ -28,6 +29,20 @@ func (r repository) create(u *model.User) error {
 	}
 
 	return err
+}
+
+func (r repository) findAll() ([]*model.User, error) {
+	var users []*model.User
+
+	err := r.db.
+		Preload("Groups").
+		Preload("AdminGroups").
+		Find(&users).Error
+	if err != nil {
+		return nil, fmt.Errorf("failed to find all users: %v", err)
+	}
+
+	return users, nil
 }
 
 func (r repository) findByEmail(email string) (*model.User, error) {
--- a/swagger/swagger.yaml
+++ b/swagger/swagger.yaml
@@ -1395,6 +1395,27 @@ paths:
             security:
                 - oauth2: []
             summary: Sign out
+        get:
+            description: Find all users with the groups they belong to
+            operationId: findAllUsers
+            responses:
+                "200":
+                    description: User
+                    schema:
+                        items:
+                            $ref: '#/definitions/User'
+                        type: array
+                "401":
+                    $ref: '#/responses/Error'
+                "403":
+                    $ref: '#/responses/Error'
+                "404":
+                    $ref: '#/responses/Error'
+                "415":
+                    $ref: '#/responses/Error'
+            security:
+                - oauth2: []
+            summary: Find users
         post:
             description: Sign up a user. This endpoint is publicly accessible and therefor anyone can sign up. However, before being able to perform any actions, users needs to be a member of a group. And only administrators can add users to groups.
             operationId: signUp
--- a/cmd/serve/main.go
+++ b/cmd/serve/main.go
@@ -142,7 +142,7 @@ func run() error {
 	r := server.GetEngine(cfg.BasePath)
 
 	group.Routes(r, authentication, authorization, groupHandler)
-	user.Routes(r, authentication, userHandler)
+	user.Routes(r, authentication, authorization, userHandler)
 	stack.Routes(r, authentication, stackHandler)
 	integration.Routes(r, authentication, integrationHandler)
 	database.Routes(r, authentication, databaseHandler)
--- a/pkg/user/router.go
+++ b/pkg/user/router.go
@@ -2,10 +2,11 @@ package user
 
 import (
 	"github.com/dhis2-sre/im-manager/internal/handler"
+	"github.com/dhis2-sre/im-manager/internal/middleware"
 	"github.com/gin-gonic/gin"
 )
 
-func Routes(r *gin.Engine, authenticationMiddleware handler.AuthenticationMiddleware, handler Handler) {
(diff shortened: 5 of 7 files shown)


pkg/user/repository.go
```diff
@@ -32,6 +32,17 @@ func (r repository) create(u *model.User) error {
 	return err
 }
 
+func (r repository) findUsers() ([]*model.User, error) {
```
teleivo, line 35: I am not sure if I would call it `findUsers` as we are getting all users. I would use `find` when we have some property/attribute we use to find a record. Not sure if `getAll`/`getAllUsers` would be better 🤔

swagger/swagger.yaml
```diff
@@ -1370,6 +1370,27 @@ paths:
             security:
                 - oauth2: []
             summary: Sign out
+        get:
+            description: List all users with the groups they belong to
```
teleivo, line 1374: I would try to keep our vocabulary to a minimum and avoid synonyms. So I would stick to `Get all` instead of `list`

teleivo (commented):
LGTM 😄 

The only things maybe worth discussing are naming.
Another things which might not be a concern now
* pagination: we don't have a lot of users now so are we saying we will introduce this once it becomes an issue
* casing: casing in our JSON is not consistent. If we want to use camelCase we should stick to it at least for everything new thats added. And fix casing for all existing responses at some point.

tonsV2:
> * pagination: we don't have a lot of users now so are we saying we will introduce this once it becomes an issue

I don't mind adding pagination to some of our endpoints if someone really wants to do so now. However I'm more in favor of dealing with it once it's an issue.

> * casing: casing in our JSON is not consistent. If we want to use camelCase we should stick to it at least for everything new thats added. And fix casing for all existing responses at some point.

Yes, we should go over all of our response types. Maybe even implement some DTO pattern.

pkg/user/repository.go
radnov: I think that both `find` and `get` can both entail finding/getting something by a property, so I went with `find` as it's more commonly used in the code base (actually, there's hardly any `getSomething` methods). I.e. I'm slightly in favour of `findAll`. What do you think?

radnov:
> > * pagination: we don't have a lot of users now so are we saying we will introduce this once it becomes an issue
> 
> I don't mind adding pagination to some of our endpoints if someone really wants to do so now. However I'm more in favor of dealing with it once it's an issue.
> 
> > * casing: casing in our JSON is not consistent. If we want to use camelCase we should stick to it at least for everything new thats added. And fix casing for all existing responses at some point.
> 
> Yes, we should go over all of our response types. Maybe even implement some DTO pattern.

@teleivo, @tonsV2 I'm also leaning toward dealing with pagination once it becomes a necessity, but I also don't mind implementing it. 👍 

About casing - I'm not sure I'm following. Are we talking about `clusterConfiguration` compared to the rest which are `CapitalCase`? But in any case, I do agree that all responses should be consistent in terms of naming and casing. 👍

teleivo:
> About casing - I'm not sure I'm following. Are we talking about `clusterConfiguration` compared to the rest which are `CapitalCase`? But in any case, I do agree that all responses should be consistent in terms of naming and casing. +1

Yes 😄 JSON property names should have consistent casing (they don't right now). If we agree on camelCase this PR should already adhere to it IMHO 😋

pkg/user/repository.go
tonsV2: > About casing - I'm not sure I'm following. Are we talking about `clusterConfiguration` compared to the rest which are `CapitalCase`? But in any case, I do agree that all responses should be consistent in terms of naming and casing. +1

It's not just `clusterConfiguration` which is lower case but that aside. Last time I investigated I didn't find a way to globally define this. So we have to add struct tags to every property we return.

tonsV2 (approved):

pkg/user/repository.go
teleivo: find and get have slightly different meaning to me but I might be wrong 😁 find: search for something that might not be there, get something I know exists. `findAll` does make it more clear at least. I'll let you decide on `findAll/getAll` 😋

pkg/user/handler.go
```diff
@@ -242,3 +243,29 @@ func (h Handler) FindById(c *gin.Context) {
 
 	c.JSON(http.StatusOK, userWithGroups)
 }
+
+// List user
```
teleivo, line 247: Now that you chose `GetAll` we should replace `List` with it everywhere 😄

pkg/user/repository.go
tonsV2: @radnov, @teleivo - I've been using findAll elsewhere. Let's stick with that or change it globally :pray:

pkg/user/handler.go
tonsV2: Please note [this](https://github.com/dhis2-sre/im-manager/pull/253#discussion_r1213896666) comment.

radnov: I finally went with `FindAll` 😄

radnov:
The PR branch is now in sync with `master` after https://github.com/dhis2-sre/im-manager/pull/251 and https://github.com/dhis2-sre/im-manager/pull/259 were merged.

pkg/user/repository.go
```diff
@@ -30,6 +30,17 @@ func (r repository) create(u *model.User) error {
...
 
+func (r repository) findAll() ([]*model.User, error) {
+	var u []*model.User
+
+	err := r.db.
+		Preload("Groups").
+		Preload("AdminGroups").
+		Find(&u).Error
+
+	return u, err
```
teleivo, line 41: I think we should not return the gorm error but something like `fmt.Errorf("failed to find all users: %v", err)` in case of an error. I don't think we agreed on something yet so we should probably do that to return consistent errors.

Like what phrase to use `failed to...` and whether to return the original error to the user (like with `%v` or not and log the details only for us).

radnov: Okay, that makes sense. Did you mean something like:
```
func (r repository) findAll() ([]*model.User, error) {
	var u []*model.User

	err := r.db.
		Preload("Groups").
		Preload("AdminGroups").
		Find(&u).Error
	if err != nil {
		log.Printf("failed to find all users: %v", err)
		return nil, fmt.Errorf("failed to find all users: %v", err)
	}

	return u, err
}
```

🤔

teleivo: yes 😄 the last return can then be `return users, nil`

PR #257, opened 2023-06-01
feat: implement deleting users by their id

radnov:
This PR adds an admin user endpoint `DELETE /users/{id}` for deleting users by their id.

I've added an [extra check to not allow the user to delete themself](https://github.com/dhis2-sre/im-manager/pull/257/files#diff-ca91d4575c0561bc512e314d455708400a1e715c5c98a109cd79960b29ee8a3aR307-R311), which I did by mistake while testing and thought it might be nice to have. Not completely sure if it's okay to have it in the handler, though. Let me know what you think. 👍

### DIFF
--- a/pkg/user/handler.go
+++ b/pkg/user/handler.go
@@ -294,6 +294,7 @@ func (h Handler) Delete(c *gin.Context) {
 
 	user, err := handler.GetUserFromContext(c)
 	if err != nil {
+		_ = c.Error(err)
 		return
 	}
 
--- a/pkg/user/handler.go
+++ b/pkg/user/handler.go
@@ -92,6 +92,7 @@ func (h *Handler) SignIn(c *gin.Context) {
 	//   415: Error
 	user, err := handler.GetUserFromContext(c)
 	if err != nil {
+		_ = c.Error(err)
 		return
 	}
 
@@ -170,6 +171,7 @@ func (h Handler) Me(c *gin.Context) {
 	//   415: Error
 	user, err := handler.GetUserFromContext(c)
 	if err != nil {
+		_ = c.Error(err)
 		return
 	}
 
@@ -199,6 +201,7 @@ func (h Handler) SignOut(c *gin.Context) {
 	//	415: Error
 	user, err := handler.GetUserFromContext(c)
 	if err != nil {
+		_ = c.Error(err)
 		return
 	}
 
--- a/pkg/user/handler.go
+++ b/pkg/user/handler.go
@@ -6,8 +6,8 @@ import (
 	"github.com/dhis2-sre/im-manager/pkg/model"
 
 	"github.com/dhis2-sre/im-manager/internal/errdef"
-
 	"github.com/dhis2-sre/im-manager/internal/handler"
+
 	"github.com/dhis2-sre/im-manager/pkg/config"
 	"github.com/dhis2-sre/im-manager/pkg/token"
 	"github.com/gin-gonic/gin"
@@ -32,6 +32,7 @@ type userService interface {
 	SignIn(email string, password string) (*model.User, error)
 	FindById(id uint) (*model.User, error)
 	FindAll() ([]*model.User, error)
+	Delete(id uint) error
 }
 
 type tokenService interface {
@@ -264,4 +265,51 @@ func (h Handler) FindAll(c *gin.Context) {
 	}
 
 	c.JSON(http.StatusOK, users)
+}
+
+// Delete user
+func (h Handler) Delete(c *gin.Context) {
+	// swagger:route DELETE /users/{id} deleteUser
+	//
+	// Delete user
+	//
+	// Delete user by id
+	//
+	// Security:
+	//	oauth2:
+	//
+	// Responses:
+	//	202:
+	//	401: Error
+	//	403: Error
+	//	404: Error
+	//	415: Error
+	id, ok := handler.GetPathParameter(c, "id")
+	if !ok {
+		return
+	}
+
+	user, err := handler.GetUserFromContext(c)
+	if err != nil {
+		return
+	}
+
+	_, err = h.userService.FindById(id)
+	if err != nil {
+		_ = c.Error(err)
+		return
+	}
+
+	if user.ID == id {
+		_ = c.Error(errdef.NewBadRequest("cannot delete the current user"))
+		return
+	}
+
+	err = h.userService.Delete(id)
+	if err != nil {
+		_ = c.Error(err)
+		return
+	}
+
+	c.Status(http.StatusAccepted)
 }
--- a/pkg/user/repository.go
+++ b/pkg/user/repository.go
@@ -75,3 +75,14 @@ func (r repository) findById(id uint) (*model.User, error) {
 	}
 	return u, err
 }
+
+func (r repository) delete(id uint) error {
+	db := r.db.Unscoped().Delete(&model.User{}, id)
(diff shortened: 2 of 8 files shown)


tonsV2 (approved):
LGTM! The only thing I'm wondering, do we really want to response with forbidden if it's the current user? Would bad request be more appropriate?

radnov:
> LGTM! The only thing I'm wondering, do we really want to respond with forbidden if it's the current user? Wouldn't bad request be more appropriate?

Yeah, to be honest I wasn't sure which one is more appropriate, [403 Forbidden](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403) or [400 Bad Request](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400). 🤷

tonsV2:
> > LGTM! The only thing I'm wondering, do we really want to respond with forbidden if it's the current user? Wouldn't bad request be more appropriate?
> 
> @tonsV2 Yeah, to be honest I wasn't sure which one is more appropriate, [403 Forbidden](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403) or [400 Bad Request](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400). shrug

I'm in favor of bad request :-)

pkg/user/service.go
```diff
@@ -124,3 +130,12 @@ func (s service) FindOrCreate(email string, password string) (*model.User, error
 
 	return s.repository.findOrCreate(user)
 }
+
+func (s service) Delete(id uint) error {
+	_, err := s.repository.findById(id)
```
teleivo, line 135: We are making 2 separate operations on the DB while one would do it. For example
 https://stackoverflow.com/questions/67154864/how-to-handle-gorm-error-at-delete-function

pkg/user/handler.go
```diff
@@ -242,3 +245,80 @@ func (h Handler) FindById(c *gin.Context) {
...
+	// Security:
+	//	oauth2:
+	//
+	// Responses:
+	//	202:
+	//	401: Error
+	//	403: Error
+	//	404: Error
+	//	415: Error
+	idParam := c.Param("id")
```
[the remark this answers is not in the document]
radnov: As #251 is not ready for merging yet (confirmed with @tonsV2), this change will have to be applied with it.

radnov:
The PR branch is now in sync with `master` after https://github.com/dhis2-sre/im-manager/pull/251 and https://github.com/dhis2-sre/im-manager/pull/259 were merged.

pkg/user/handler.go
```diff
@@ -242,3 +245,80 @@ func (h Handler) FindById(c *gin.Context) {
...
+	id, err := strconv.ParseUint(idParam, 10, 32)
+	if err != nil {
+		badRequest := apperror.NewBadRequest("error parsing id")
+		_ = c.Error(badRequest)
+		return
+	}
+
+	user, err := handler.GetUserFromContext(c)
+	if err != nil {
+		return
```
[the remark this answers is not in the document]
teleivo: It seems like we always do `_ = c.Error(err)` on error when calling `GetUserFromContext`.

@tonsV2 should we change the signature to one like `GetPathParameter` so that we write

```go
	user, ok := handler.GetUserFromContext(c)
	if ok != nil {
       return
    }
```

pkg/user/handler.go
```diff
@@ -265,3 +269,52 @@ func (h Handler) FindAll(c *gin.Context) {
...
+
+	_, err = h.userService.FindById(uint(id))
+	if err != nil {
+		_ = c.Error(err)
+		return
+	}
+
+	if user.ID == uint(id) {
+		forbiddenRequest := errdef.NewBadRequest("cannot delete the current user")
+		_ = c.Error(forbiddenRequest)
```
teleivo, line 309: if you wanted to return a https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403
you would need to do

`c.AbortWithError(http.StatusForbidden, errors.New("cannot delete the current user"))`

pkg/user/repository.go
```diff
@@ -75,3 +75,14 @@ func (r repository) findById(id uint) (*model.User, error) {
 	}
 	return u, err
 }
+
+func (r repository) delete(id uint) error {
+	db := r.db.Unscoped().Delete(&model.User{}, id)
+	if db.Error != nil {
+		return db.Error
```
teleivo, line 82: Same as in https://github.com/dhis2-sre/im-manager/pull/253#discussion_r1221514101 applies. To me the rule is if the error does not originate from us (errors.New, fmt.Errorf) then we need to turn it into ours by calling `fmt.Errorf` or `errdef`. This way errors will be understandable for our users 😄

pkg/user/handler.go
radnov: I actually forgot to update the error variable name after changing the status code to Bad Request, as also suggested by @tonsV2 [here](https://github.com/dhis2-sre/im-manager/pull/257#issuecomment-1573155335). 😅

radnov:
> It seems like we always do `_ = c.Error(err)` on error when calling `GetUserFromContext`.
> 
> @tonsV2 should we change the signature to one like `GetPathParameter` so that we write
> 
> ```go
> 	user, ok := handler.GetUserFromContext(c)
> 	if ok != nil {
>        return
>     }
> ```

That sounds good to me. 👍  (_p.s. GitHub's PR comment "replies" are something else ..._ 😞)

pkg/user/handler.go
```diff
@@ -265,3 +270,51 @@ func (h Handler) FindAll(c *gin.Context) {
...
+	}
+
+	_, err = h.userService.FindById(uint(id))
+	if err != nil {
+		_ = c.Error(err)
+		return
+	}
+
+	if user.ID == uint(id) {
+		_ = c.AbortWithError(http.StatusBadRequest, errors.New("cannot delete the current user"))
```
teleivo, line 309: That's not what I meant 😅 

If you are fine with the status codes in https://github.com/dhis2-sre/im-manager/blob/9060e0af8ba4eaeae67309f45b7b4f29d8248ba3/internal/middleware/errorHandler.go#L27 then all you need is `_ = c.Error(err)`. So if you want HTTP status code 400 do `_ = c.Error(errdef.NewBadRequest("cannot delete the current user"))`. `errdef` IMHO contains and should only contain HTTP independent/generic errors used throughout the app. The mapping to HTTP happens in the HTTP aware error middleware `errorHandler`.

If you want to return a specific HTTP status code that we do not already map in the `errorHandler` use `_ = c.AbortWithError(http.StatusForbidden, errors.New("cannot delete the current user"))` with the status code in a handler.

teleivo:
> LGTM! The only thing I'm wondering, do we really want to respond with forbidden if it's the current user? Wouldn't bad request be more appropriate?

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403

> This status is similar to [401](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401), but for the 403 Forbidden status code, re-authenticating makes no difference. The access is tied to the application logic, such as insufficient rights to a resource.

403 sounds appropriate to me, but I am also ok with 400

pkg/user/handler.go
radnov: Okay, I get it now. I was slightly confused why I had to use `AbortWithError` and just went with your previous comment. 😅

teleivo (approved):

PR #259, opened 2023-06-02
chore: remove apperror

tonsV2:
The main intention of this PR is to get rid of apperror. For the most this hasn't affected the code but I did make various minor changes for the sake of making the code more idiomatic and cleaner. Such changes include handling "not found" errors in the repositories only returning points instead of actual objects a few places.


internal/handler/dataBinder.go
```diff
@@ -3,19 +3,20 @@ package handler
...
 	if c.ContentType() != "application/json" && c.ContentType() != "multipart/form-data" {
 		reason := fmt.Sprintf("%s only accepts content of type application/json or multipart/form-data", c.FullPath())
-		return apperror.NewUnsupportedMediaType(reason)
+		return errdef.NewUnsupportedMediaType(reason)
 	}
 
 	if err := c.ShouldBind(req); err != nil {
 		message := fmt.Sprintf("Error binding data: %+v\n", err)
-		return apperror.NewBadRequest(message)
+		return errdef.NewBadRequest(message)
```
teleivo, line 19:
```suggestion
		return errdef.NewBadRequest("Error binding data: %+v\n", err)
```

should work (if so it should be applied everywhere).

internal/handler/dataBinder.go
```diff
@@ -3,19 +3,20 @@ package handler
...
+	"github.com/dhis2-sre/im-manager/internal/errdef"
+
 	"github.com/gin-gonic/gin"
 )
 
 func DataBinder(c *gin.Context, req interface{}) error {
 	if c.ContentType() != "application/json" && c.ContentType() != "multipart/form-data" {
 		reason := fmt.Sprintf("%s only accepts content of type application/json or multipart/form-data", c.FullPath())
-		return apperror.NewUnsupportedMediaType(reason)
+		return errdef.NewUnsupportedMediaType(reason)
```
teleivo, line 14: I would follow the [im-user internal/handler/dataBinder.go](https://github.com/dhis2-sre/im-user/pull/16/files#diff-81b269c1e73a24de897a1f4e7831d0428479c171e871dfe1f9489f2deae6c4f5) and the guideline in https://github.com/dhis2-sre/im-user/pull/16

`errdef` to me should have HTTP independent errors. The HTTP error middleware than turns these into appropriate HTTP status codes.

pkg/database/repository.go
```diff
@@ -45,6 +51,9 @@ func (r repository) FindBySlug(slug string) (*model.Database, error) {
 		Preload("Lock").
 		Where("slug = ?", slug).
 		First(&d).Error
+	if errors.Is(err, gorm.ErrRecordNotFound) {
+		return nil, errdef.NewNotFound("database not found by slug: %s", slug)
```
teleivo, line 55: I would quote strings using `%q`. This string might not contain any whitespace but its probably an easier convention than to decide based on what string it is 🤔

internal/middleware/errorHandler.go
```diff
@@ -1,25 +1,44 @@
...
+		if c.Writer.Status() != http.StatusOK {
+			_, _ = c.Writer.WriteString(err.Error())
+			return
+		}
+
+		// nolint:gocritic
+		if errdef.IsBadRequest(err) {
+			c.String(http.StatusBadRequest, err.Error())
+		} else if errdef.IsUnsupportedMediaType(err) {
+			c.String(http.StatusBadRequest, err.Error())
```
radnov, line 30: Should be `http.StatusUnsupportedMediaType`

pkg/database/repository.go
```diff
@@ -57,31 +66,33 @@ func (r repository) Lock(id, instanceId, userId uint) (*model.Lock, error) {
...
 		if err != nil {
+			if errors.Is(err, gorm.ErrRecordNotFound) {
+				err = errdef.NewNotFound("database not found by id: %d", id)
+			}
 			return err
 		}
 
 		if d.Lock != nil && d.Lock.InstanceID != 0 {
-			return fmt.Errorf("database already locked by user \"%d\" and instance \"%d\"", userId, d.Lock.InstanceID)
+			return errdef.NewBadRequest("database already locked by user \"%d\" and instance \"%d\"", userId, d.Lock.InstanceID)
```
teleivo, line 76: use `%q` to quote

internal/middleware/errorHandler.go
```diff
@@ -1,25 +1,44 @@
...
+		} else if errdef.IsDuplicated(err) {
+			c.String(http.StatusBadRequest, err.Error())
```
radnov, line 32: Should this be `http.StatusConflict`? 🤔

pkg/database/repository.go
```diff
@@ -57,31 +66,33 @@ func (r repository) Lock(id, instanceId, userId uint) (*model.Lock, error) {
...
+		return tx.Create(lock).Error
 	})
 
 	return lock, errTx
 }
 
 func (r repository) Unlock(id uint) error {
-	return r.db.Delete(&model.Lock{}, id).Error
+	err := r.db.Delete(&model.Lock{}, id).Error
+	if errors.Is(err, gorm.ErrRecordNotFound) {
```
teleivo, line 92: Gorm delete does not seem to return ErrRecordNotFound if what we are trying to delete does not exist https://stackoverflow.com/questions/67154864/how-to-handle-gorm-error-at-delete-function

radnov (commented):
Looks good, but seems like there's some mismatches in the error handler middleware.

pkg/database/repository.go
```diff
@@ -114,16 +125,19 @@ func (r repository) CreateExternalDownload(databaseID uint, expiration time.Time
...
 
-func (r repository) FindExternalDownload(uuid uuid.UUID) (model.ExternalDownload, error) {
-	var d model.ExternalDownload
+func (r repository) FindExternalDownload(uuid uuid.UUID) (*model.ExternalDownload, error) {
+	var d *model.ExternalDownload
 	err := r.db.
 		Where("expiration > ?", time.Now().UTC()).
 		First(&d, uuid).Error
+	if errors.Is(err, gorm.ErrRecordNotFound) {
+		return nil, errdef.NewNotFound("external download not found by id: %s", uuid.String())
```
teleivo, line 134:
```suggestion
		return nil, errdef.NewNotFound("external download not found by id: %q", uuid)
```

pkg/database/service.go
```diff
@@ -193,7 +150,7 @@ func (s service) Download(id uint, dst io.Writer, cb func(contentLength int64))
 	}
 
 	if d.Url == "" {
-		return apperror.NewBadRequest(fmt.Sprintf("database with %d doesn't reference any url", id))
+		return errdef.NewBadRequest("database with %d doesn't reference any url", id)
```
teleivo, line 153:
```suggestion
		return errdef.NewBadRequest("database with id %d doesn't reference any url", id)
```

might not be clear what `database with %d` means

pkg/instance/handler.go
```diff
@@ -209,16 +210,16 @@ func (h Handler) consumeParameters(user *model.User, sourceInstanceId uint, inst
 	}
 
 	if preset && !sourceInstance.Preset {
-		return apperror.NewUnauthorized(fmt.Sprintf("instance (id: %d) isn't a preset", sourceInstance.ID))
+		return errdef.NewUnauthorized(fmt.Sprintf("instance (id: %d) isn't a preset", sourceInstance.ID))
```
teleivo, line 213: I don't think you need to call `fmt.Sprintf` as it will delegate to `fmt.Errorf` anyway

internal/handler/dataBinder.go
tonsV2: Shouldn't we remove \n?

pkg/database/repository.go
tonsV2: Good catch!

pkg/database/repository.go
```diff
@@ -57,31 +66,38 @@ func (r repository) Lock(id, instanceId, userId uint) (*model.Lock, error) {
...
 }
 
 func (r repository) Unlock(id uint) error {
-	return r.db.Delete(&model.Lock{}, id).Error
+	db := r.db.Delete(&model.Lock{}, id)
+	if db.Error != nil {
+		return db.Error
+	}
+
+	if db.RowsAffected < 1 {
```
teleivo, line 96: @radnov made `Delete` return `NewNotFound` when deleting a user that does not exist. We should be consistent. If we want that behavior here as well we should move that logic into repository.Delete and call that method from here 😄

pkg/instance/handler.go
```diff
@@ -569,31 +570,23 @@ func (h Handler) Delete(c *gin.Context) {
 		return
 	}
 
-	token, err := handler.GetTokenFromHttpAuthHeader(c)
-	if err != nil {
-		_ = c.Error(err)
-		return
-	}
-
 	instance, err := h.instanceService.FindById(uint(id))
```
teleivo, line 573: Aren't these already returning a NotFound error? In that case we can simplify the if (and similar ones in any handler) to

```go
	instance, err := h.instanceService.FindById(uint(id))
    if err != nil {
	    _ = c.Error(err)
	    return
    }
```

radnov (approved):

internal/handler/dataBinder.go
```diff
@@ -3,19 +3,20 @@ package handler
...
-		return apperror.NewUnsupportedMediaType(reason)
+		message := fmt.Sprintf("%s only accepts content of type application/json or multipart/form-data", c.FullPath())
+		return errdef.NewBadRequest(message)
 	}
 
 	if err := c.ShouldBind(req); err != nil {
-		message := fmt.Sprintf("Error binding data: %+v\n", err)
-		return apperror.NewBadRequest(message)
+		message := fmt.Sprintf("Error binding data: %+v", err)
+		return errdef.NewBadRequest(message)
```
teleivo, line 19: I think you overlooked what I actually meant, unless I was wrong 😋. I think the call to fmt.Sprintf is redundant and should be removed 😄 

```suggestion
		return errdef.NewBadRequest("Error binding data: %+v", err)
```

pkg/database/repository.go
tonsV2: But this is about deleting a lock, not a user?

internal/handler/dataBinder.go
tonsV2: Yeah, the fmt.Sprintf should be removed but further more I think it make sense to remove the `UnsupportedMediaType` error type as well. By the end of the day, not posting the right content type is a bad request.

teleivo: yes which is what I already advocated for in comment https://github.com/dhis2-sre/im-manager/pull/259/files/3ded1ea09b64d95b2592c706a165a349916ce02d#r1214145209 the only thing thats left is removing the redundant fmt.Sprintf call

pkg/database/repository.go
teleivo: my bad I probably mistaken this as the service 🤦🏻 

should we use `r.db.Unscoped().Delete` as we mostly seem to do. the `r.db.Delete(&model.Lock{}, id)` does a soft-delete AFAIK

https://github.com/dhis2-sre/im-manager/pull/257/files#diff-c7bd001c6323abf762d3175df7f3e7c28aaa063386901005ee236254b39b007dR80

internal/handler/dataBinder.go
tonsV2: Already done, I'm not sure why it's not reflected here

pkg/database/repository.go
tonsV2: Already pushed, I'm not sure why it's not reflected here :man_shrugging:

PR #260, opened 2023-06-02
fix: save

tonsV2:
The "Save as" function for database has been failing for some time. I assume it's due to an update applied by dependabot but haven't bothered investigating.

This PR fixes the issue but it's worth nothing that the code dealing with saving of databases could be improved. More details can be found [here](https://github.com/aws/aws-sdk-go/issues/1962).


radnov (approved):
Nice! 👍

teleivo (approved):
I am confused that the sdk issue is closed and this seems to be the fix 🤷🏻

PR #263, opened 2023-06-07
Feat/env

tonsV2:
The goal of this PR is to get rid of the user specific environments. It can be tested by adding CLASSIFICATION=dev to your .env file and keep ENVIRONMENT as your own environment.

I tried to install the roles using skaffold profiles but when I run `skaffold dev --trigger=manual -p dev` it doesn't install the im-manager itself. Are you guys also having this issue?

The above can be circumvented by running both `skaffold dev --trigger=manual -p dev` and `skaffold dev --trigger=manual` but I was under the impression that it should be needed.


radnov:
@tonsV2 `skaffold dev --trigger=manual -p dev` indeed only deploys the `im-group` chart. 

This seems to be because using profiles will override the whole `deploy` config, as described [here](https://skaffold.dev/docs/environment/profiles/#profiles-profiles), as far as I understand it. Seems like `skaffold` can't combine the multiple `helm.releases` from both the main `skaffold.yaml` and `profiles` config.

skaffold.yaml
```diff
@@ -41,35 +41,56 @@ deploy:
...
         namespace: instance-manager-{{ .ENVIRONMENT }}
         createNamespace: true
         chartPath: helm/chart
         setValueTemplates:
-          image.repository: dhis2/im-manager
+          image:
+            repository: dhis2/im-manager
+          serviceAccount:
+            annotations:
+              eks\\.amazonaws\\.com\\/role-arn: arn:aws:iam::767224633206:role/instance-cluster-test-im-manager-{{ .ENVIRONMENT }}
```
radnov, line 56: The key here needs to be quoted, otherwise I'm getting the following error when running `skaffold run/dev`

```
Error: UPGRADE FAILED: YAML parse error on im-manager/templates/serviceaccount.yaml: error unmarshaling JSON: while decoding JSON: json: cannot unmarshal object into Go struct field .metadata.annotations of type string
```

radnov (commented):
`skaffold dev --trigger=manual -p dev` indeed only deploys the `im-group` chart. This is because using profiles will override the whole `deploy` config, as described [here](https://skaffold.dev/docs/environment/profiles/#profiles-profiles), as far as I understand it. Seems like `skaffold` can't combine the multiple `helm.releases` from both the main `skaffold.yaml` and `profiles` config.


p.s. PR looks good in general, but the branch needs to be synced with master to do a full review.

radnov (commented):
Looks good overall, but maybe we should also move away from the user specific stack parameters. Currently we have Google Earth Engine credentials for both `dhis2` and `dhis2-core` and the DB manager URL for `dhis2` and `dhis2-db` (which might even be redundant at this point, or at least needs renaming?).

Also updating the [`.sops.yaml` config file](https://github.com/dhis2-sre/im-manager/blob/feat/env/.sops.yaml) with a simpler regex for the paths after the changes (i.e. only `prod` and `dev` subdirs) would be nice. Maybe even updating the `non-prod` keys aliases 🤔

radnov (approved):
Approving, as [my previous comment and suggestions](https://github.com/dhis2-sre/im-manager/pull/263#pullrequestreview-1476616384) are not required as part of this PR, so we don't have to stall this one.

PR #264, opened 2023-06-08
feat: implement updating users' email and/or password

radnov:
This PR implements and endpoint for updating a user's email and/or password by their ID.

### DIFF
--- a/pkg/user/handler.go
+++ b/pkg/user/handler.go
@@ -33,6 +33,7 @@ type userService interface {
 	FindById(id uint) (*model.User, error)
 	FindAll() ([]*model.User, error)
 	Delete(id uint) error
+	Update(id uint, email, password string) (*model.User, error)
 }
 
 type tokenService interface {
@@ -316,4 +317,51 @@ func (h Handler) Delete(c *gin.Context) {
 	}
 
 	c.Status(http.StatusAccepted)
+}
+
+type UpdateUserRequest struct {
+	Email    string `json:"email" binding:"omitempty,email"`
+	Password string `json:"password" binding:"omitempty,gte=16,lte=128"`
+}
+
+// Update user
+func (h *Handler) Update(c *gin.Context) {
+	// swagger:route PUT /users/{id} updateUser
+	//
+	// Update user
+	//
+	// Update user's email and/or password
+	//
+	// security:
+	//   oauth2:
+	//
+	// responses:
+	//   200: User
+	//   401: Error
+	//   403: Error
+	//   404: Error
+	//   415: Error
+	id, ok := handler.GetPathParameter(c, "id")
+	if !ok {
+		return
+	}
+
+	var request UpdateUserRequest
+	if err := handler.DataBinder(c, &request); err != nil {
+		_ = c.Error(err)
+		return
+	}
+
+	if request.Email == "" && request.Password == "" {
+		_ = c.Error(errdef.NewBadRequest("neither email nor password are specified"))
+		return
+	}
+
+	user, err := h.userService.Update(id, request.Email, request.Password)
+	if err != nil {
+		_ = c.Error(err)
+		return
+	}
+
+	c.JSON(http.StatusOK, user)
 }
--- a/pkg/user/repository.go
+++ b/pkg/user/repository.go
@@ -86,3 +86,17 @@ func (r repository) delete(id uint) error {
 
 	return nil
 }
+
+func (r repository) update(user *model.User) (*model.User, error) {
+	updatedUser := model.User{
+		Email:    user.Email,
+		Password: user.Password,
+	}
+
+	err := r.db.Model(&user).Updates(updatedUser).Error
+	if err != nil {
+		return nil, fmt.Errorf("failed to update user: %v", err)
+	}
+
+	return user, nil
+}
--- a/pkg/user/service.go
+++ b/pkg/user/service.go
@@ -23,6 +23,7 @@ type userRepository interface {
 	findOrCreate(email *model.User) (*model.User, error)
 	findAll() ([]*model.User, error)
 	delete(id uint) error
+	update(user *model.User) (*model.User, error)
 }
 
 type service struct {
@@ -120,7 +121,7 @@ func (s service) FindById(id uint) (*model.User, error) {
 func (s service) FindOrCreate(email string, password string) (*model.User, error) {
 	hashedPassword, err := hashPassword(password)
 	if err != nil {
-		return nil, fmt.Errorf("password hashing failed: %s", err)
+		return nil, fmt.Errorf("failed to hash password: %s", err)
 	}
 
 	user := &model.User{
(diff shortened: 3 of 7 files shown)


pkg/user/handler.go
```diff
@@ -265,3 +267,56 @@ func (h Handler) FindAll(c *gin.Context) {
...
+		return
+	}
+
+	err := h.userService.Update(uint(id), request.Email, request.Password)
+	if err != nil {
+		_ = c.Error(err)
+		return
+	}
+
+	user, err := h.userService.FindById(uint(id))
```
tonsV2, line 315: You don't need the type conversion

pkg/user/handler.go
```diff
@@ -265,3 +267,56 @@ func (h Handler) FindAll(c *gin.Context) {
...
+	}
+
+	var request UpdateUserRequest
+	if err := handler.DataBinder(c, &request); err != nil {
+		_ = c.Error(err)
+		return
+	}
+
+	if request.Email == "" && request.Password == "" {
+		_ = c.AbortWithError(http.StatusBadRequest, errors.New("neither email nor password are specified"))
```
tonsV2, line 305: you can use errdef.NewBadRequest here

pkg/user/service.go
```diff
@@ -129,3 +130,24 @@ func (s service) FindOrCreate(email string, password string) (*model.User, error
...
+
+	if password != "" {
+		var err error
+		user.Password, err = hashPassword(password)
+		if err != nil {
+			return fmt.Errorf("password hashing failed: %s", err)
+		}
+	}
+
+	return s.repository.update(user)
```
tonsV2, line 152: Is the user object updated when calling the update method? I'm wondering if we might as well return it? You're reloading it in the handler

scripts/users/update.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+echo "{
+  \"email\": \"$USER_EMAIL\",
+  \"password\": \"$PASSWORD\"
+}" | $HTTP put "$IM_HOST/users/$1" "Authorization: Bearer $ACCESS_TOKEN"
+
```
tonsV2, line 9: Do we really want all of this here?

pkg/user/handler.go
```diff
@@ -265,3 +267,56 @@ func (h Handler) FindAll(c *gin.Context) {
...
+	Password string `json:"password" binding:"omitempty,gte=16,lte=128"`
+}
+
+// Update user
+func (h *Handler) Update(c *gin.Context) {
+	// swagger:route PUT /users/{id} updateUser
+	//
+	// Update user
+	//
+	// Update user email and/or password
```
tonsV2, line 282: user's

pkg/user/handler.go
```diff
@@ -265,3 +267,56 @@ func (h Handler) FindAll(c *gin.Context) {
...
+		_ = c.Error(err)
+		return
+	}
+
+	if request.Email == "" && request.Password == "" {
+		_ = c.AbortWithError(http.StatusBadRequest, errors.New("neither email nor password are specified"))
+		return
+	}
+
+	err := h.userService.Update(uint(id), request.Email, request.Password)
```
tonsV2, line 309: You don't need the type conversion

tonsV2 (changes requested):

scripts/users/update.sh
radnov: Do you mean the email and password JSON body for the request? I just went with the simplest example of updating both. 🤔

pkg/user/service.go
radnov: Yep, seems like this is the better way to go about it. 👍

pkg/user/handler.go
```diff
@@ -317,3 +318,50 @@ func (h Handler) Delete(c *gin.Context) {
...
+	//   404: Error
+	//   415: Error
+	id, ok := handler.GetPathParameter(c, "id")
+	if !ok {
+		return
+	}
+
+	var request UpdateUserRequest
+	if err := handler.DataBinder(c, &request); err != nil {
+		_ = c.Error(err)
```
teleivo, line 351: @tonsV2 @radnov there seems to be a pattern of handler methods that process user input. When the processing fails we should stop processing the request setting the error into the Gin context. For example `GetPathParameter` (which you adapted), `GetUserFromContext` see https://github.com/dhis2-sre/im-manager/pull/257#discussion_r1223792766 and `DataBinder` as seen here.

We could adapt such methods in a way similar to `GetPathParameter` so here it would be

```go
if !handler.DataBinder(c, &request) {
  return
}
```

with `DataBinder` setting the error into the Gin context returning false on error. The downside of such signatures is that with returning `error` its clearer that we should stop processing than it with with a method returning a `bool`. Ignoring a returned `error` is also caught by linting while `bool` is not. Whatever we choose we should be consistent. Thinking through this I actually favor returning `error` 🤔

pkg/user/repository.go
```diff
@@ -86,3 +86,14 @@ func (r repository) delete(id uint) error {
...
 	return nil
 }
+
+func (r repository) update(user *model.User) (*model.User, error) {
+	updatedUser := model.User{
+		Email:    user.Email,
+		Password: user.Password,
+	}
+
+	err := r.db.Model(&user).Updates(updatedUser).Error
```
teleivo, line 96: https://github.com/dhis2-sre/im-doc/pull/4/files#diff-e14025c0fa40d4857e4b40fc96ea5ee995afe300014626b68cb55a479fa5b8fcR49

pkg/user/service.go
```diff
@@ -134,3 +135,24 @@ func (s service) FindOrCreate(email string, password string) (*model.User, error
...
+
+	if email != "" {
+		user.Email = email
+	}
+
+	if password != "" {
+		var err error
+		user.Password, err = hashPassword(password)
+		if err != nil {
+			return nil, fmt.Errorf("password hashing failed: %s", err)
```
teleivo, line 153: nit: as mentioned in another comment we should settle on a standard phrase like `failed to ...`/`error ...`. Like `failed to hash password` or `error hashing password`

```suggestion
			return nil, fmt.Errorf("failed to hash password: %s", err)
```

pkg/user/handler.go
radnov: I'm also in favour of returning an `error` instead of `bool`. 👍  I suppose adapting the `GetUserFromContext` and `DataBinder` methods to be as `GetPathParameter` should be done in a seaprate PR, though?

pkg/user/repository.go
radnov: 🤦‍♂️

pkg/user/service.go
radnov: Also changed this in another place within the same file.

pkg/user/handler.go
teleivo: If we are in favor of returning an `error` than we don't need to change `DataBinder` and `GetUserFromContext`. The intention of `GetPathParameter` was to remove repetition, not have to do the type conversion to uint and set the correct HTTP status code https://github.com/dhis2-sre/im-manager/blob/f3e6a2dc4d4b05c45592c6c782bde8656ac57fe1/internal/handler/pathParameter.go#L15. Due to the latter there is no need to return an error as its already in the Gin context. Lets leave this for now.

teleivo (approved):

tonsV2 (approved):

PR #276, opened 2023-06-14
refactor: align all JSON responses to be in camelCase

radnov:
This PR aligns all JSON API response keys to be in `camelCase`. The embedded `gorm.Model` is removed from all our models and instead the necessary fields are just copied. Note that I left the `DeletedAt` field, even though we're not using it anywhere right now, but maybe it's better to remove it.

You can see the [updated API docs](https://api.im.radnov.test.c.dhis2.org/docs) on my environment.


teleivo:
`Time` and `Valid` are returned for `deletedAt`. If we decide to remove `deletedAt` thats fine (I vote for removing it 😄). If not we need to lower case them.

Then there is fields that adhere to https://github.com/golang/go/wiki/CodeReviewComments#initialisms which we should camelCase as well
* UUID
* databaseID (and maybe other fields like it) 

https://api.im.radnov.test.c.dhis2.org/docs#operation/listDatabases
```json
{
  "createdAt": "2019-08-24T14:15:22Z",
  "deletedAt": {
    "Time": "2019-08-24T14:15:22Z",
    "Valid": true
  },
  "externalDownloads": [
    {
      "UUID": "f50af7e0-0dd5-4361-ab96-2e04f7bc7e30",
      "databaseID": 0,
      "expiration": "2019-08-24T14:15:22Z"
    }
  ],
  "groupName": "string",
  "id": 0,
  "lock": {
    "databaseID": 0,
    "instanceID": 0,
    "userID": 0
  },
  "name": "string",
  "slug": "string",
  "updatedAt": "2019-08-24T14:15:22Z",
  "url": "string"
}
```

Can we also please merge this after https://github.com/dhis2-sre/im-manager/pull/275/files 😅 I think it would be great to have some more tests in first.

radnov:
@teleivo I didn't bother messing with the `deletedAt` object fields, as I think we should get rid of it. @tonsV2 Do you agree with removing it?

I did check the [initialisms and acronyms guideline](https://github.com/golang/go/wiki/CodeReviewComments#initialisms) and that's why I chose `databaseID` instead of `databaseId` (same goes for `userID`, `instanceID`, etc), isn't that correct?
Now `UUID` could indeed be `uuid` I suppose, given that it's an acronym and we'll be using "consistent case" for it.  




> Can we also please merge this after #275

Yep, of course. 👍

teleivo:
> @teleivo I didn't bother messing with the `deletedAt` object fields, as I think we should get rid of it. @tonsV2 Do you agree with removing it?
> 
> I did check the [initialisms and acronyms guideline](https://github.com/golang/go/wiki/CodeReviewComments#initialisms) and that's why I chose `databaseID` instead of `databaseId` (same goes for `userID`, `instanceID`, etc), isn't that correct? Do you mean that we should use `databaseid`, etc? I think that kind of breaks the `camelCase` consistency more. thinking Now `UUID` could indeed be `uuid` I suppose, given that it's an acronym and we'll be using "consistent case" for it.

I would do

* `userID` in Go and `userId` in JSON
* `UUID` in Go and `uuid` in JSON

basically putting whatever we have in Go through an algorithm that converts it to camelCase. An algorithm that does not know anything about acronyms or English 😂

pkg/model/group.go
```diff
@@ -11,18 +11,20 @@ const AdministratorGroupName = "administrators"
...
-	Users                []User                `gorm:"many2many:user_groups;constraint:OnUpdate:CASCADE,OnDelete:CASCADE;"`
+	Name                 string                `gorm:"primarykey; unique;" json:"name"`
+	Hostname             string                `gorm:"unique;" json:"hostname"`
+	Users                []User                `gorm:"many2many:user_groups;constraint:OnUpdate:CASCADE,OnDelete:CASCADE;" json:"users"`
 	ClusterConfiguration *ClusterConfiguration `json:"clusterConfiguration"`
-
-	CreatedAt time.Time
-	UpdatedAt time.Time
-	DeletedAt gorm.DeletedAt `gorm:"index"`
+	CreatedAt            time.Time             `json:"createdAt"`
```
teleivo, line 18: nit: we could have the fields that came from gorm.Model always at the top

pkg/model/stack.go
```diff
@@ -36,14 +34,14 @@ func (s Stack) FindOptionalParameter(name string) (StackOptionalParameter, error
 }
 
 type StackRequiredParameter struct {
-	Name      string `gorm:"primaryKey"`
-	StackName string `gorm:"primaryKey"`
-	Consumed  bool
+	Name      string `gorm:"primaryKey" json:"name"`
```
teleivo, line 37: These are done automatically by the `json.Marshaler` so no json tag would be needed. We can stick to being consistent and still add them like you did though 👌🏻 😄

scripts/databases/auth.sh
```diff
@@ -8,6 +8,6 @@ exp=$(echo "$ACCESS_TOKEN" | jq -R 'split(".")? | .[1] | @base64d | fromjson | .
 NOW=$(date +%s)
 if [[ -z "$exp" ]] || (( $exp < $NOW )); then
   # shellcheck disable=SC2155
-  export ACCESS_TOKEN=$($HTTP --auth "$USER_EMAIL:$PASSWORD" post "$IM_HOST/tokens" | jq -r '.access_token')
+  export ACCESS_TOKEN=$($HTTP --auth "$USER_EMAIL:$PASSWORD" post "$IM_HOST/tokens" | jq -r '.accessToken')
```
tonsV2, line 11: Why? :smile:

tonsV2 (changes requested):

scripts/databases/auth.sh
radnov: Because the `Token`  field is also updated - https://github.com/dhis2-sre/im-manager/pull/276/files#diff-73f556a7b2309746d2b616a31bf69f24de0334f4707ce826f3ca185feadbef24R49-R52

tonsV2: As mentioned on Slack... The reason for it being of that format is that I wanted to mimic the spring security implementation.

https://docs.spring.io/spring-security/oauth/apidocs/org/springframework/security/oauth2/common/OAuth2AccessToken.html

The motivation were spring being widely used and therefor a lot of 3rd party integrations would work out of the box. But I guess we don't need to integrate with many random 3rd party applications.

Another thing worth keeping in mind. This update will break interaction im-web-client. So that needs to be updated as well.

tonsV2 (approved):

PR #284, opened 2023-06-16
test(wip): test  instance handler

teleivo:
based on https://github.com/dhis2-sre/im-manager/pull/275

The [last commit](https://github.com/dhis2-sre/im-manager/pull/284/commits/1af9fd0e7fe4a4aa6f8ea840f91606126ed787f8) in this PR adds an integration test for the instance handler using k3s.

The test is deploying `whoami-go`. It is not complete yet! Whats missing is
- [ ] either configure sops or mock it. Maybe https://pkg.go.dev/github.com/orlangure/gnomock/preset/localstack#section-readme has what we need. Right now I set the KUBECONFIG in the `helmfile.go` which should come from the groups cluster config. Setting that is easy, the difficulty is that its not encrypted using sops while that is what is expected AFAIK
- [ ] assert that `whoami-go` is up. It would be amazing if we could make an HTTP GET to it. Otherwise, we can use the kubernetes client and assert the pod is up.
- [ ] if all of the above is working then extract a helper into `inttest`. See comment and other helpers for the style 😄 
- [ ] In https://github.com/dhis2-sre/im-manager/pull/275 I said we should usually test packages in isolation. I did not isolate the instance test from the stack package. Mostly because I had not much time but also because I wanted to deploy `whoami-go`. We can keep it that way or mock the stack service and return an even simpler stack than whoami-go or return `whoami-go`. Important is that we don't have too many tests failing for the same reason.


teleivo:
Closing for now as the code has change a lot/will change and I am out of time right now. Anyone feel free to resurrect this 😄

PR #306, opened 2023-06-26
refactor: improve databases seed script

radnov:
As we start to get users actually using the Instance Manager, we need to have the databases frequently seeded and kept in sync. This PR improves the databases `seed.sh` to allow for easier seeding of the most frequently used databases, as well as provide an option for specifying DB versions manually.

The seed script will now have two options for usage:
* With only a group name provided `./seed whoami` - this will default to seeding databases for all DHIS2 versions (taken from the [stable.json](https://releases.dhis2.org/v1/versions/stable.json) file) for the current and 2 previous versions, without hotfixes, as we don't publish specific database versions for hotfixes.
* With group name and explicit database versions provided `./seed whoami 2.40 2.39`.

### DIFF
--- a/scripts/databases/seed.sh
+++ b/scripts/databases/seed.sh
@@ -3,23 +3,43 @@
 set -euo pipefail
 
 GROUP=$1
+shift
+VERSIONS=$*
+
+DHIS2_RELEASES_URL="https://releases.dhis2.org/v1/versions/stable.json"
 
 function createDatabase() {
-  echo "Downloading database... $1"
-  curl -C - "$2" -o "$HOME/Downloads/$1.sql.gz"
-  #curl "$2" -o "$HOME/Downloads/$1.sql.gz"
+  echo "Downloading database $1 ..."
 
-  echo "Login..."
+  mkdir -p "$HOME/Downloads"
+  curl -C - "$2" -o "$HOME/Downloads/$1.sql.gz"
+
+  echo "Login ..."
+  rm .access_token_cache # to make sure we're not using an expired token if we're seeding a lot of databases
   source ./auth.sh
 
-  echo "Uploading database...$1"
+  echo "Uploading database $1 ..."
   ./upload.sh "$GROUP" "$HOME/Downloads/$1.sql.gz"
+  echo # empty line to improve output readability
 }
 
-## https://databases.dhis2.org/
-createDatabase "Sierra Leone - 2.39.0" https://databases.dhis2.org/sierra-leone/2.39.0/dhis2-db-sierra-leone.sql.gz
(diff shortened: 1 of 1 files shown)


scripts/databases/seed.sh
```diff
@@ -3,23 +3,44 @@
 set -euo pipefail
 
 GROUP=$1
+shift
+DATABASE_VERSIONS=$*
```
tonsV2, line 7: Kinda nit pick but maybe remove the "DATABASE_" prefix. It should be clear from the context but I really don't feel strong about this one. It just came to mind when I saw [this](https://github.com/dhis2-sre/im-manager/pull/306/files#diff-5f9562bf7fb13c1dedb12edd181ae94501e089f2b309987a684e2be742a86f4bR42) line.

scripts/databases/seed.sh
```diff
@@ -3,23 +3,44 @@
...
+  DATABASE_VERSIONS+=($(
+    echo "$dhis2_versions" |
+    jq -r "[.versions[] |
+    select(.version == $latest_version or .version == ($latest_version - 1) or .version == ($latest_version - 2)) |
+    [.name] + [(.patchVersions[] | select(.hotfix != true) | .name)]] | flatten | .[]"
+  ))
+fi
+
+echo "Seeding the following databases:"
+echo "${DATABASE_VERSIONS[@]}"
```
tonsV2, line 40:
```suggestion
echo "Seeding the following databases: ${DATABASE_VERSIONS[@]}"
```

scripts/databases/seed.sh
```diff
@@ -3,23 +3,44 @@
...
+
+for VERSION in "${DATABASE_VERSIONS[@]}"; do
+  createDatabase "Sierra Leone - $VERSION" "https://databases.dhis2.org/sierra-leone/$VERSION/dhis2-db-sierra-leone.sql.gz"
+done
 
-./list.sh | jq '.[].Databases[]? | .ID, .Name, .Url'
+./list.sh | jq '.[].databases[]? | .id, .name, .url'
```
tonsV2, line 46: I'm wondering if the piping to jq actually should be part of the script? It makes it a lot less flexible.

Another problem is that it lists for all groups regardless of the input so it can be a bit misleading too

tonsV2 (approved):

scripts/databases/seed.sh
radnov: It's not really needed, yes. I'll go ahead and remove it. 👍

scripts/databases/seed.sh
radnov: I split it into two lines because of https://github.com/koalaman/shellcheck/wiki/SC2145, but I'll just disable the check for that line. 👍

PR #311, opened 2023-06-29
feat: deployable/non-deployable groups

tonsV2:
We want to prevent users deploying to certain groups such as "administrators" and some "database only" groups


radnov (approved):
Very nice! Looks good to me 👍

PR #314, opened 2023-07-05
feat: public instances

tonsV2:
The goal of this PR is to introduce a method to mark an instance as public and an endpoint which will return all public instances


radnov (approved):
Looks good to me 👌

PR #318, opened 2023-07-06
refactor: run instance-manager workflow on PR for feature branches

radnov:
This PR enables running the updated [Build, test and deploy workflow](https://github.com/dhis2-sre/gha-workflows/pull/29) on PRs and only deploy when the `deploy` label is present (either initially added when creating the PR or after the PR was already created), as well as adds a new [Delete environment](https://github.com/dhis2-sre/gha-workflows/pull/29/files) that deletes a deployed environment when the PR is closed/merged. Note that the behaviour of the "Build, test and deploy" workflow remains the same for pushes to `master`, as well as creating new tags.

Details: 
* The `ENVIRONMENT` and `CLASSIFICATION` variables are used to distinguish between the separate "feature" environments that are deployed within the same `instance-manager-feature` k8s namespace.
* `helm/data/secrets/feature` and `helm/data/values/feature` are symlinks to the respective `dev` environment directories.
* Disable concurrently running multiple runs of the same workflow - only the latest triggered run for the given workflow and PR (or git ref) will run and all previous runs will be canceled.
* `skaffold` profiles are updated to use [patches](https://skaffold.dev/docs/environment/profiles/#override-via-patches) to override (or in this case add to) the main config with the appropriate `im-group` deployments for the given environment, which allows us to run a single `skaffold` command (like `skaffold run -p dev`), in order to deploy all needed resources for an environment. To get the "effective" configuration that will be executed with a profile applied try running `skaffold diagnose -p dev/prod`.
* The `skaffold.yaml` file now has two YAML anchors to remove the code duplications in the profiles.
* Deploying PR branches is triggered by adding a `deploy` label to the PR (either initially or after the PR is already created).
* Deployed feature environments are deleted when the PR is closed/merged.
* Creating a PR without a `deploy-feature` label will only run the build and test parts of the workflow, without deploying. This should be useful for PRs that we don't want to deploy to the cluster, but still want to test (like dependabot ones).

I've tested deploying multiple PR environments within the same namespace with [this test PR](https://github.com/dhis2-sre/im-manager/pull/323) and verified that uploading a database, as well as deploying a `dhis2` stack from both environments works as expected. 

I've also tested the workflow for deleting the deployed environment with [this PR](https://github.com/dhis2-sre/im-manager/pull/333), which when closed triggered [this run](https://github.com/dhis2-sre/im-manager/actions/runs/5541001477).


skaffold.yaml
```diff
@@ -81,64 +81,75 @@ deploy:
...
-            version: 0.3.0
-            setValueTemplates:
-              serviceAccount:
-                name: im-manager-{{ .ENVIRONMENT }}
-                namespace: instance-manager-{{ .ENVIRONMENT }}
+    patches:
+      - op: add
+        path: /deploy/helm/releases/-
+        value:
+          name: im-group-whoami-{{ .ENVIRONMENT }}
```
tonsV2, line 88: Since we're only changing name and namespace here, we've a lot of duplicated code. Could we use a yaml anchor to simplify things?

tonsV2 (approved):

tonsV2 (approved):

.github/workflows/delete-env.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+# Cancel previous runs of the same workflow and PR number or branch/tag
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  call-workflow:
+    # TODO update the branch reference before merging the PR
+    uses: dhis2-sre/gha-workflows/.github/workflows/im-delete-env.yaml@DEVOPS-262
```
teleivo, line 15: Is it true that right now we always run `skaffold delete` on closing a PR other than a dependabot PR? Are you doing this so that when someone adds the deploy label and removes it again we still clean up when the PR is closed?

skaffold.yaml
```diff
@@ -1,4 +1,16 @@
-apiVersion: skaffold/v4beta2
+.im-group: &chart-config
```
teleivo, line 1: Why is the key different from the anchor?

Especially with `.im-group-prod: &service-account-config` we loose that the anchor references prod values. This is a bit confusing to me 🤔

skaffold.yaml
```diff
@@ -81,64 +93,45 @@ deploy:
...
-              serviceAccount:
-                name: im-manager-prod
-                namespace: instance-manager-prod
+    patches:
+      - op: add
+        path: /deploy/helm/releases/-
+        value:
+          name: im-group-dev
+          namespace: dev
+          # TODO use YAML anchor
```
teleivo, line 115: I think you finished the TODO 😄

skaffold.yaml
```diff
@@ -81,64 +93,45 @@ deploy:
...
-            version: 0.3.0
-            setValues:
-              serviceAccount:
-                name: im-manager-prod
-                namespace: instance-manager-prod
+    patches:
+      - op: add
+        path: /deploy/helm/releases/-
+        value:
+          name: im-group-dev
```
teleivo, line 113: Why do we use `dev` in name/namespace in prod?

skaffold.yaml
```diff
@@ -81,64 +93,45 @@ deploy:
...
+          namespace: dev
+          # TODO use YAML anchor
+          <<: [*chart-config, *service-account-config]
 
-          - name: im-group-play
```
teleivo, line 113: Did you remove `- name: im-group-play` and if so intentionally? It looks as if the same happened for `im-group-qa` and `im-group-meta-packages` 🤔

.github/workflows/delete-env.yaml
radnov: Yep, exactly right. That seemed like the easiest option for making sure to clean up after a PR regardless of whether it was just closed, merged and whether something actually got deployed, etc.

skaffold.yaml
radnov: Maybe slightly ambiguous, but that's for the `prod` environment to be able to deploy to the `dev` namespace.

skaffold.yaml
radnov: The diff is slightly misleading here. Look at the added lines below or the actual contents of the file after the changes.

skaffold.yaml
radnov: Sure, I can make the key and anchor the same, just looked weird to me. 🤷

skaffold.yaml
teleivo: Interesting 😅 why are we doing that?

radnov: Because we want the `im-manager` deployed to the `instance-manager-prod` environment to be able to deploy to the `dev` namespace which is assigned to the `dev` group. If it's just the naming again, do you have any better alternatives?

teleivo (approved):

PR #324, opened 2023-07-11
feat: human-readable error messages when signing up

tonsV2:


radnov (approved):
Looks good to me. Note that I have not tested it locally, though.

pkg/user/user_integration_test.go
```diff
@@ -65,13 +65,53 @@ func TestUserHandler(t *testing.T) {
...
 
+	t.Run("SignUpFailure", func(t *testing.T) {
+		t.Parallel()
+
+		t.Run("InvalidEmail", func(t *testing.T) {
+			requestBody := strings.NewReader(`{
+				"email":    "not-a-valid-email",
+				"password": "oneoneoneoneoneoneone111"
+			}`)
+			response := client.Do(t, http.MethodPost, "/users", requestBody, http.StatusBadRequest, inttest.WithHeader("Content-Type", "application/json"))
```
teleivo, line 83: The tests are layed out in AAA arrange (the setup), act (the things that is being tested), assert so far. To be consistent I would add newlines between `requestBody`, `response` and the `require` 😄

pkg/user/user_integration_test.go
```diff
@@ -65,13 +65,53 @@ func TestUserHandler(t *testing.T) {
...
+				"password": "short-password"
+			}`)
+			response := client.Do(t, http.MethodPost, "/users", requestBody, http.StatusBadRequest, inttest.WithHeader("Content-Type", "application/json"))
+			require.Equal(t, "password must be between 24 and 128 characters", string(response))
+		})
+
+		t.Run("TooLongPassword", func(t *testing.T) {
+			requestBody := strings.NewReader(`{
+				"email":    "some@email.com",
+				"password": "long-password-long-password-long-password-long-password-long-password-long-password-long-password-long-password-long-password-long-password"
```
teleivo, line 99: You could use `strings.Repeat("a", 129)` which makes it easy to spot why the password is invalid.

pkg/user/user_integration_test.go
```diff
@@ -65,13 +65,53 @@ func TestUserHandler(t *testing.T) {
...
 			"email":    "user2@dhis2.org",
-			"password": "oneoneoneoneone1"
+			"password": "oneoneoneoneoneoneone111"
 		}`), &user2)
 
 		require.Equal(t, "user2@dhis2.org", user2.Email)
 		require.Empty(t, user2.Password)
 	}
 
+	t.Run("SignUpFailure", func(t *testing.T) {
```
teleivo, line 75: `SignUpFailed` would be more consistent with the naming of the existing test `SignInFailed`

pkg/user/handler.go
```diff
@@ -72,6 +72,27 @@ func (h Handler) SignUp(c *gin.Context) {
 	c.JSON(http.StatusCreated, user)
 }
 
+func handleSignUpErrors(err error) error {
+	validationErrors, ok := err.(validator.ValidationErrors)
+	if ok {
```
teleivo, line 77: I would invert the condition to reduce nesting. So `if !ok { return errdef.NewBadRequest("Error binding data: %+v", err) }` then the field error logic 😄

teleivo (commented):
Nice 😄 

We spoke about it sometime to maybe return JSON instead of a plain string. Now that we are returning multiple errors. On the other hand maybe the frontend could already validate these fields and highlight the inputs that are invalid.

PR #341, opened 2023-07-19
fix: remove "required" binding from Deployable bool field

radnov:
Turns out that if the Deployable bool field is `false`, the validation from the "required" tag will fail
```
Error #01: Key: 'CreateGroupRequest.Deployable' Error:Field validation for 'Deployable' failed on the 'required' tag
```
as it's considered the zero-value for bools. In our case both `true` and `false` are valid, so there's no need to use the "required" tag.

I've also handled the duplicated key error when creating a `group`, but I'm not sure how to distinguish between duplicated `Hostname` and `Name`. Any ideas?

### DIFF
--- a/pkg/group/repository.go
+++ b/pkg/group/repository.go
@@ -83,7 +83,13 @@ func findAllFromUser(user *model.User, deployable bool) []model.Group {
 }
 
 func (r repository) create(group *model.Group) error {
-	return r.db.Create(&group).Error
+	err := r.db.Create(&group).Error
+	if errors.Is(err, gorm.ErrDuplicatedKey) {
+		// TODO how to check if name or hostname is duplicated?
+		return errdef.NewDuplicated("group name/hostname already exists: %s", err)
+	}
+
+	return err
 }
 
 func (r repository) findOrCreate(group *model.Group) (*model.Group, error) {
--- a/pkg/group/docs.go
+++ b/pkg/group/docs.go
@@ -2,7 +2,7 @@ package group
 
 // swagger:parameters groupCreate
 type _ struct {
-	// Refresh token request body parameter
+	// Create group request body parameter
 	// in: body
 	// required: true
 	Body CreateGroupRequest
--- a/pkg/group/handler.go
+++ b/pkg/group/handler.go
@@ -33,7 +33,7 @@ type groupService interface {
 type CreateGroupRequest struct {
 	Name       string `json:"name" binding:"required"`
 	Hostname   string `json:"hostname" binding:"required"`
-	Deployable bool   `json:"deployable" binding:"required"`
+	Deployable bool   `json:"deployable"`
 }
 
 // Create group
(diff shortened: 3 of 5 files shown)


pkg/group/repository.go
```diff
@@ -83,7 +83,13 @@ func findAllFromUser(user *model.User, deployable bool) []model.Group {
 }
 
 func (r repository) create(group *model.Group) error {
-	return r.db.Create(&group).Error
+	err := r.db.Create(&group).Error
+	if errors.Is(err, gorm.ErrDuplicatedKey) {
+		// TODO how to check if name or hostname is duplicated?
```
tonsV2, line 88: That's a good question! Since it's just the uniq constraint which is failing I don't think we'll know unless the error message from Postgresql shows some details

tonsV2 (approved):

teleivo (approved):
If it was not for https://github.com/dhis2-sre/im-manager/pull/342 I would request a test 😋 so you are off the hook 😁

pkg/group/repository.go
radnov: Unfortunately, it seems like with the [TranslateError](https://github.com/dhis2-sre/im-manager/blob/master/pkg/storage/postgresql.go#L24) flag enabled for `Gorm` the original Postgresql error details are gone and you only get ["duplicated key not allowed"](https://github.com/go-gorm/gorm/blob/master/errors.go#L49). 

Without that flag here's how the original error looks like in JSON:
```
{
  "Severity": "ERROR",
  "Code": "23505",
  "Message": "duplicate key value violates unique constraint \"groups_pkey\"",
  "Detail": "Key (name)=(sl2) already exists.",
  "Hint": "",
  "Position": 0,
  "InternalPosition": 0,
  "InternalQuery": "",
  "Where": "",
  "SchemaName": "public",
  "TableName": "groups",
  "ColumnName": "",
  "DataTypeName": "",
  "ConstraintName": "groups_pkey",
  "File": "nbtinsert.c",
  "Line": 670,
  "Routine": "_bt_check_unique"
}
```

But without the TranslateFlag we have to go back to checking the error `Code` to handle errors. 🤔

radnov: @teleivo ☝️ this probably applies for [your comment](https://github.com/dhis2-sre/im-manager/pull/341#discussion_r1267959256), as well

teleivo: Do we have control over the constraint name? If both constraints have a clear discernible name we could check `"ConstraintName": "groups_pkey"` right?

radnov: With [TranslateError](https://github.com/dhis2-sre/im-manager/blob/master/pkg/storage/postgresql.go#L24) enabled (which is needed in order to check for `gorm.ErrDuplicatedKey`) we don't have control over the contstraint name, but without we do. This seems to be confirmed by [this](https://github.com/go-gorm/gorm/issues/6327#issuecomment-1555048172) and the next comment.

teleivo: `Name                 string                `json:"name" gorm:"primarykey; unique;"` turns into constraint name `groups_pkey`

and what is the constraint called?

`Hostname             string                `json:"hostname" gorm:"unique;"`

We can than just use a best-effort and check that the `"ConstraintName"` of the error is either an exact match or contains some substring.

PR #342, opened 2023-07-19
feat: remove user from a group and return a group with its users

tonsV2:
The goal of this PR is to implement some of the endpoints missing for building a user management on the client. Specifically it implements functionality for removing a user from a group and retrieving a group and its users.

Since there were no tests for the groups handler a few extra have been added. I also corrected a few typos so Intellij is reporting less warnings.


pkg/inttest/http.go
```diff
@@ -109,15 +110,15 @@ func (hc *HTTPClient) Do(t *testing.T, method, path string, requestBody io.Reade
 	return body
 }
 
-// do delegates the request to the underlying HTTP client.
+// do delegate the request to the underlying HTTP client.
```
teleivo, line 113: That comment was correct 😅  IMHO as the function `do` is the subject as in `the function do delegates the request`

Same applies to the change in the `Do` methods comment

pkg/inttest/http.go
```diff
@@ -109,15 +110,15 @@ func (hc *HTTPClient) Do(t *testing.T, method, path string, requestBody io.Reade
...
 func (hc *HTTPClient) do(t *testing.T, req *http.Request) *http.Response {
 	resp, err := hc.Client.Do(req)
 	require.NoError(t, err, httpClientErrMessage(req.Method, req.URL.Path)+": HTTP request failed")
 	return resp
 }
 
 // GetJSON sends an HTTP GET request to given path. Optional headers are applied to the request. The
-// response body is unmarshaled as JSON into given responseBody. Failure to read or close the HTTP
+// response body is unmarshalled as JSON into given responseBody. Failure to read or close the HTTP
```
teleivo, line 121: https://pkg.go.dev/encoding/json uses `marshaled`

Seems like you are changing it to British English
https://www.oxfordlearnersdictionaries.com/definition/english/marshal_2?q=marshaled
I would suggest we stick to US English but am ok not to, at least we should stick to one of the two 😅

pkg/group/group_integration_test.go
```diff
@@ -0,0 +1,196 @@
...
+			require.Equal(t, "non-deployable-test-group", group.Name)
+			require.Equal(t, "non-deployable-test-hostname.com", group.Hostname)
+			require.Equal(t, false, group.Deployable)
+		})
+	})
+
+	t.Run("AddUserToGroup", func(t *testing.T) {
+		t.Parallel()
+
+		t.Run("AddUserToGroup", func(t *testing.T) {
```
teleivo, line 97: You are repeating the parents subtest name. We could call it

```suggestion
		t.Run("Success", func(t *testing.T) {
```

like we do in the user intttest. This can also serve as an easy to remember filter for running the test alone.

Same for the other subtests. The names are concatenated in `subtest1/subtest2/...`. I would therefore not repeat the parents name but try to make it read more like a sentence.

pkg/group/group_integration_test.go
```diff
@@ -0,0 +1,196 @@
...
+
+			require.Equal(t, "failed to find user with id 123", string(response))
+		})
+	})
+
+	t.Run("RemoveUserFromGroup", func(t *testing.T) {
+		t.Parallel()
+
+		t.Run("RemoveUserFromGroup", func(t *testing.T) {
+			err := groupService.AddUser(groupName, user.ID)
```
teleivo, line 127: If I am not mistaken we are adding the same user to the same group `AddUserToGroup` is adding the user to. If that is true, we could run into a race condition as we remove the same user from the group in here.

We could maybe
1. rearrange the tests to write them more like a flow of create user, add user to group, get user, get user with details, remove user.
2. Or we use distinct users/groups so such races cannot occur.

I favor 1. as we are more consistent in testing via HTTP, then sometimes via HTTP and sometimes via a service layer.

pkg/group/group_integration_test.go
```diff
@@ -0,0 +1,196 @@
...
+		t.Run("RemoveUserFromNonExistingGroup", func(t *testing.T) {
+			path := fmt.Sprintf("/groups/%s/users/%s", "non-existing-group", userId)
+
+			response := client.Do(t, http.MethodPost, path, nil, http.StatusNotFound)
+
+			require.Equal(t, "group \"non-existing-group\" doesn't exist", string(response))
+		})
+
+		t.Run("RemoveNonExistingUserFromGroup", func(t *testing.T) {
+			nonExistingUserId := uint(123)
```
teleivo, line 143: We discussed this a while back and you did try it so you are not against it completely 😋 I would try to inline such vars as much as possible so the code becomes more succinct. I don't mind repetition in tests, but favor less clutter.

```go
response := client.Do(t, http.MethodDelete, "/groups/testgroup/99999", nil, http.StatusNotFound)

require.Equal(t, "failed to find user with id 99999", string(response))
```

As we assert on 404 and the `failed to find user` we are also certain that this is the cause. So there is no need to do the `userService.FindById` + `require.Error` check.

pkg/group/group_integration_test.go
```diff
@@ -0,0 +1,196 @@
...
+			var group model.Group
+			client.GetJSON(t, path, &group)
+
+			require.Equal(t, groupName, group.Name)
+		})
+
+		t.Run("FindGroupFailed", func(t *testing.T) {
+			path := fmt.Sprintf("/groups/%s", "non-existing-group")
+
+			client.Do(t, http.MethodGet, path, nil, http.StatusNotFound)
```
teleivo, line 169:
```suggestion
			client.Do(t, http.MethodGet, "/groups/non-existing-group", nil, http.StatusNotFound)
```

pkg/group/repository.go
```diff
@@ -33,6 +33,21 @@ func (r repository) find(name string) (*model.Group, error) {
...
+		Where("name = ?", name).
+		Joins("ClusterConfiguration").
+		Preload("Users").
+		Preload("AdminUsers").
+		First(&group).Error
+	if errors.Is(err, gorm.ErrRecordNotFound) {
+		return nil, errdef.NewNotFound("group %q doesn't exist", name)
+	}
+
+	return group, err
```
teleivo, line 48: I know we are not 100% consistent yet but we did say that we were going to return our own errors instead of 3rd party library ones. Here you are returning a gorm error that will then directly get passed to a user https://github.com/dhis2-sre/im-doc/blob/master/dev/README.md#general

teleivo (commented):
Great work adding these tests 🎉 

With the `get groups with users` change I think user/group are more justified to become one package 😁

pkg/group/group_integration_test.go
```diff
@@ -0,0 +1,196 @@
...
+		t.Run("Deployable", func(t *testing.T) {
+			requestBody := strings.NewReader(`{
+				"name": "deployable-test-group",
+				"hostname": "deployable-test-hostname.com",
+				"deployable": true
+			}`)
+
+			var group model.Group
+			client.PostJSON(t, "/groups", requestBody, &group)
+			require.Equal(t, "deployable-test-group", group.Name)
```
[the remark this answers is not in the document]
tonsV2: If the first one is wrong, it's quite likely that the following are wrong as well. Which is why I'm just bailing early. Would you change this elsewhere? Or just for this test?

pkg/group/group_integration_test.go
```diff
@@ -0,0 +1,196 @@
...
+	var user *model.User
+	{
+		user, err = userService.FindOrCreate("user@dhis2.org", "oneoneoneoneoneoneone111")
+		require.NoError(t, err)
+		userId = strconv.FormatUint(uint64(user.ID), 10)
+	}
+
+	var groupName string
+	{
+		group, err := groupService.Create("test-group", "test-hostname.com", true)
```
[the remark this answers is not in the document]
tonsV2: I'm using the service because it's simpler.

I'm confused about what you mean. Would you use the service in `NonDeployable` instead?

radnov (approved):
Looks good to me 👍

pkg/group/group_integration_test.go
teleivo: You are in a way testing `groupService.AddUser(groupName, user.ID)` even if its not your intention.

As said I would

> Rearrange the tests to write them more like a flow of create user, add user to group, get user, get user with details, remove user.

Maybe https://github.com/dhis2-sre/im-manager/blob/b60455c6d6253d393febac1b368931a7a42101d3/pkg/user/user_integration_test.go#L179-L207 helps as an example.

- In a `AsAdmin/Success` test (these tests could be structured as subtests of `TestGroupHandler`). Then in blocks not subtests as these should be executed sequentially
- Create `user1`
- Create group `x`
- Add `user1` to a group `x`
- Get group `x`
- Get group `x` with users and assert `user1` is in it
- Remove `user1` from the group `x`
- Delete group `x`

This simulates a flow that users might actually go through 😄 The flow will create its own user and group. This way there cannot be a race between different tests trying to add/delete the same user to and from a group.

pkg/group/repository.go
```diff
@@ -23,13 +24,32 @@ func NewRepository(db *gorm.DB) *repository {
...
-		Preload("ClusterConfiguration").
+		Joins("ClusterConfiguration").
 		Where("name = ?", name).
 		First(&group).Error
 	if errors.Is(err, gorm.ErrRecordNotFound) {
 		return nil, errdef.NewNotFound("group %q doesn't exist", name)
 	}
 
+	if err != nil {
+		return nil, fmt.Errorf("failed to find group with details: %v", err)
```
teleivo, line 35: Is this `find(name)` method also getting the group with details? Or should this error message go into `findWithDetails(name string)` 🤔

pkg/group/group_integration_test.go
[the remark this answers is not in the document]
tonsV2: I was thinking that this wasn't a test. This is just setting up data for the actual test which is further down which is why I'm using the service here

pkg/model/group.go
```diff
@@ -13,6 +13,7 @@ type Group struct {
 	Hostname             string                `json:"hostname" gorm:"unique;"`
 	Deployable           bool                  `json:"deployable"`
 	Users                []User                `json:"users" gorm:"many2many:user_groups;constraint:OnUpdate:CASCADE,OnDelete:CASCADE;"`
+	AdminUsers           []User                `json:"adminUsers" gorm:"many2many:user_groups_admin;constraint:OnUpdate:CASCADE,OnDelete:CASCADE;"`
```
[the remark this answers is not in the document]
tonsV2: Because we want a details with client side showing users and admin users

teleivo (approved):

PR #364, opened 2023-07-31
feat: log user agent, ip and user email

tonsV2:


internal/server/engine.go
```diff
@@ -4,32 +4,40 @@ import (
 	"github.com/dhis2-sre/im-manager/internal/middleware"
 	"github.com/dhis2-sre/im-manager/pkg/health"
 	"github.com/gin-contrib/cors"
+	"github.com/gin-contrib/logger"
 	"github.com/gin-gonic/gin"
 	redocMiddleware "github.com/go-openapi/runtime/middleware"
+	"github.com/rs/zerolog"
```
teleivo, line 10: should we adopt https://go.dev/blog/slog in Go 1.21 instead?

tonsV2: Yes, we should :smile:

teleivo:
@tonsV2 I think this can be closed right? we log user agent, ip and user id using slog

PR #393, opened 2023-08-17
Validate email

tonsV2:
When a user signs up we'd like to send them a mail so we can confirm they actually own the email they used.

This PR implements the above and also an endpoint which allows the users to submit a token and thereby validate their email.

Currently the email simply contains a link directly to the api but once this PR is merged the intention is to create page on im-ui which post the token to the api and advise the user about the result.


pkg/user/user_integration_test.go
```diff
@@ -194,7 +226,7 @@ func TestUserHandler(t *testing.T) {
 			var users []model.User
 			client.GetJSON(t, "/users", &users, inttest.WithAuthToken(adminToken.AccessToken))
 
-			assert.Lenf(t, users, 3, "GET /users should return 3 users one of which is an admin")
+			assert.Lenf(t, users, 4, "GET /users should return 3 users one of which is an admin")
```
radnov, line 229: Should update the expected number of users in the message, too

radnov (approved):
Looks good to me 👍  Just one minor comment.

PR #417, opened 2023-08-23
refactor: move environment subdomain at the beggining of hostnames

radnov:
Related to https://github.com/dhis2-sre/dhis2-infrastructure/pull/126

Removes the `test` subdomain from all IM environments and instances hostnames, removes the `prod` subdomain for the prod environment and moves the environment subdomain for `dev` and `feature` branches at the beginning.

Note that the "whoami" group will share the same hostname across all of the environments, as it's only used for internal testing. Here's an example of a deployed instance - https://whoami.im.dhis2.org/test-deploy-02

More details in https://dhis2.atlassian.net/browse/DEVOPS-238


scripts/databases/.env.example
```diff
@@ -3,4 +3,4 @@ HTTP="http --verify=no --check-status"
 USER_EMAIL=some@user.org
 PASSWORD=<password>
 
-IM_HOST=https://api.im.feature.test.c.dhis2.org/skaffold
+IM_HOST=https://feature.api.im.dhis2.org
```
tonsV2, line 6: Shouldn't we target dev here?

scripts/instances/.env.example
```diff
@@ -1,4 +1,4 @@
 USER_EMAIL=some@user.org
 PASSWORD=<password>
 
-IM_HOST=https://api.im.feature.test.c.dhis2.org/skaffold
+IM_HOST=https://feature.api.im.dhis2.org
```
tonsV2, line 4: Same as above

scripts/users/.env.example
```diff
@@ -4,6 +4,6 @@ PASSWORD=<password>
 ADMIN_USER_EMAIL=<email>
 ADMIN_USER_PASSWORD=<password>
 
-IM_HOST=https://api.im.feature.test.c.dhis2.org/skaffold
+IM_HOST=https://feature.api.im.dhis2.org
```
tonsV2, line 7: dev?

tonsV2 (approved):

PR #445, opened 2023-09-04
refactor: update loader to assert statically defined stacks match wha…

tonsV2:
…t's defined in the helmfile's


pkg/model/stack.go
```diff
@@ -12,33 +12,31 @@ type Stack struct {
...
-	Instances        []Instance      `json:"instances" gorm:"constraint:OnUpdate:CASCADE,OnDelete:CASCADE;"`
+	Parameters       StackParameters `json:"-" gorm:"-:all"`
+	Instances        []Instance      `json:"instances" gorm:"constraint:OnUpdate:CASCADE,OnDelete:CASCADE"`
 	HostnamePattern  string          `json:"hostnamePattern"`
 	HostnameVariable string          `json:"hostnameVariable"`
 	// Providers provide parameters to other stacks.
 	Providers Providers `json:"-" gorm:"-"`
 	// Requires these stacks to deploy an instance of this stack.
-	Requires []Stack `json:"-" gorm:"-"`
+	Requires []Stack `json:"requires" gorm:"constraint:OnUpdate:CASCADE,OnDelete:CASCADE"`
```
teleivo, line 26: I would not yet expose this in our JSON and also not yet persist it until its needed. So we can refactor freely.

pkg/stack/loader.go
```diff
@@ -36,50 +34,44 @@ func LoadStacks(dir string, stackService Service) error {
...
-			continue
-		}
 
-		name := entry.Name()
-		log.Printf("Parsing stack: %q\n", name)
-		stackTemplate, err := parseStack(dir, name)
-		if err != nil {
-			return fmt.Errorf("error parsing stack %q: %v", name, err)
-		}
+	for name := range stacks {
```
teleivo, line 38: Can we not do `name, stack := range stacks` 🤔

pkg/stack/loader.go
```diff
@@ -36,50 +34,44 @@ func LoadStacks(dir string, stackService Service) error {
...
 		}
 
-		stack := &model.Stack{
-			Name:             name,
-			HostnamePattern:  stackTemplate.hostnamePattern,
-			HostnameVariable: stackTemplate.hostnameVariable,
-			Parameters:       make(map[string]model.StackParameter),
+		parsedStack, err := parseStack(dir, name)
+		if err != nil {
+			return fmt.Errorf("error parsing stack %q: %w", name, err)
```
teleivo, line 58:
```suggestion
			return fmt.Errorf("error parsing stack %q: %v", name, err)
```

We don't yet wrap errors as we don't want to expose them i.e. allow layers above to act on it. We might in the future. But if there is on condition acting on the cause we should not wrap.

pkg/stack/loader.go
```diff
@@ -36,50 +34,44 @@ func LoadStacks(dir string, stackService Service) error {
...
 		}
-		for name, v := range stackTemplate.parameters {
-			isConsumed := isConsumedParameter(name, stackTemplate.consumedParameters)
-			stack.Parameters[name] = model.StackParameter{Consumed: isConsumed, DefaultValue: v}
+
+		parsedParameters := len(maps.Keys(parsedStack.parameters))
+		stackParameters := len(stack.Parameters)
+		if parsedParameters != stackParameters {
+			return fmt.Errorf("number of stack parameters (%d) doesn't match helmfile parameters (%d) for stack: %q", stackParameters, parsedParameters, name)
```
teleivo, line 64: loader_test.go creates a diff between helmfile parameters and the ones in the stack.go file so there is no need to do this here 😄 

We could move the parsing of stacks into loader_test.go entirely.

pkg/model/stack.go
```diff
@@ -12,33 +12,31 @@ type Stack struct {
...
-	// StackParameters. Only use GormParameters within the repository. Otherwise use
-	// Parameters.
-	GormParameters []StackParameter `json:"parameters" gorm:"foreignKey:StackName; references: Name; constraint:OnUpdate:CASCADE,OnDelete:CASCADE"`
+	// GormParameters are only used by Gorm to persist parameters as it cannot persist a StackParameters.
+	// Only use GormParameters within the repository. Otherwise, use Parameters.
+	GormParameters []StackParameter `json:"parameters" gorm:"foreignKey:StackName; references:Name; constraint:OnUpdate:CASCADE,OnDelete:CASCADE"`
 	// Parameters used by the stacks helmfile template.
-	Parameters       StackParameters `json:"-" gorm:"-"`
-	Instances        []Instance      `json:"instances" gorm:"constraint:OnUpdate:CASCADE,OnDelete:CASCADE;"`
+	Parameters       StackParameters `json:"-" gorm:"-:all"`
```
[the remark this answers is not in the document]
tonsV2: >   Name string `gorm:"-:all"`        // ignore this field when write, read and migrate with struct
Source: https://gorm.io/docs/models.html

But I guess, we just want "-"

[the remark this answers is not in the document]
tonsV2: No, we want gorm to completely ignore this field! Which is why I used ":all"

pkg/stack/loader.go
tonsV2: Please note the todo comment on the next line

teleivo: let go with your solution. In 1.21 we could try to enable https://github.com/golang/go/wiki/LoopvarExperiment

pkg/model/instance.go
```diff
@@ -35,18 +35,15 @@ type Linked struct {
...
 
 type InstanceParameter struct {
-	InstanceID uint `json:"-" gorm:"primaryKey"`
-	// TODO: Rename StackParameterID to Name
-	StackParameterID string         `json:"name" gorm:"primaryKey"`
-	StackParameter   StackParameter `json:"-" gorm:"foreignKey:StackParameterID,StackName; references:Name,StackName; constraint:OnUpdate:CASCADE,OnDelete:CASCADE"`
-	StackName        string         `json:"-"`
-	Value            string         `json:"value"`
+	InstanceID         uint   `json:"-" gorm:"primaryKey"`
+	StackParameterName string `json:"name" gorm:"primaryKey"`
```
teleivo, line 47: * Since we are within the struct `InstanceParameter` we know that the field is one of a parameter.
* Since the `InstanceParameter` goes through validation we also know that it needs to match the stack the instance is associated with.

I would therefore like to keep it short and call this field `Name`. This would then also match the name in `StackParameter`.

pkg/model/stack.go
```diff
@@ -2,49 +2,19 @@ package model
...
-	UpdatedAt time.Time `json:"updatedAt"`
-	// GormParameters are only used by Gorm to persist parameters as it cannot persist a
-	// StackParameters. Only use GormParameters within the repository. Otherwise use
-	// Parameters.
-	GormParameters []StackParameter `json:"parameters" gorm:"foreignKey:StackName; references: Name; constraint:OnUpdate:CASCADE,OnDelete:CASCADE"`
-	// Parameters used by the stacks helmfile template.
-	Parameters       StackParameters `json:"-" gorm:"-"`
-	Instances        []Instance      `json:"instances" gorm:"constraint:OnUpdate:CASCADE,OnDelete:CASCADE;"`
+	Name             string          `json:"name"`
+	Parameters       StackParameters `json:"parameters"`
```
teleivo, line 10: You are introducing a breaking change here going from `[]` to a `{}` of parameters. Is that intentional? `{}` is what I prefer to have in the end 😋 just checking

pkg/stack/service.go
```diff
@@ -2,50 +2,23 @@ package stack
...
-
-	err := s.repository.CreateParameter(parameter)
-
-	return parameter, err
+type service struct {
+	stacks Stacks
 }
 
 func (s service) Find(name string) (*model.Stack, error) {
-	return s.repository.Find(name)
```
teleivo, line 42: `s.repository.Find(name)` returned `errdef.NewNotFound("stack not found by name: %s", name)` if the stack is not found. I think we should still do that, or not?

tonsV2: We definitely should

pkg/model/instance.go
tonsV2: So would I! But previously I haven't been able to make it work due to a sql error... ambiguous column name or something, I don't recall the exact error

pkg/stack/service.go
```diff
@@ -15,7 +16,10 @@ type service struct {
 }
 
 func (s service) Find(name string) (*model.Stack, error) {
```
teleivo, line 18: In `FindAll` we return a copy of the stacks `[]model.Stack`. I think we should also return `model.Stack` instead of `*model.Stack` here

teleivo (approved):

PR #446, opened 2023-09-04
chore: add parameter validation mechanism DEVOPS-223

teleivo:
* stack parameters can define a `Validator` which is a `func (value string) error`. I am not adding an interface like for `Provider`. I don't anticipate us needing any `struct` holding state like a `AWS` client or so that should satisfy this interface. Validators IMHO will mostly be functions. Functions cannot implement an interface in Go which leads to the pattern like the [http/HandlerFunc](https://pkg.go.dev/net/http#HandlerFunc) which makes things more complex. We can go there if needed.
* adapted the minimum parameter validation we had in instance service. New algorithm: collect as many errors as possible by iterating over all instance parameters
  * add error if parameter is not part of the instance's stack
  * add error if there is a Validator to run that returns an error
  * join all errors in a `BadRequest` error

Using higher-order functions we can create reusable validators like OneOf. If we benefit from adding multiple Validators to one parameter we can do so at a later point. Being able to return multiple errors in a structured way like JSON instead of a plain string might help with the presentation in the UI.

## Example Request
 
`./deploy-whoami.sh whoami foo`

with the following added invalid parameters 

```json
    {
      "name": "IMAGE_PULL_POLICY",
      "value": "nonsense"
    },
    {
      "name": "UNKNOWN",
      "value": "some"
    }
```

Parameter `IMAGE_PULL_POLICY` is invalid according to `Validator: ImagePullPolicy`. The other one is not part of the whoami stack.

Leads to HTTP 400 Bad Request with message

`invalid parameter(s): parameter "IMAGE_PULL_POLICY": "nonsense" is not valid, only "Always", "Never", "IfNotPresent" are allowed
parameter "UNKNOWN": not a stack parameter`

## Next steps for validation

* add more Validators for each stack parameter
* validatiors are run in the `instanceService.Save` which is used in the instance deploy and update handler. Are there other handlers that take in instance parameters that need to be validated?
* think of how to improve the error representation in the UI. Should we move to a structured error response like JSON where we return a list of errors maybe associated with each parameter?

### DIFF
--- a/pkg/instance/service.go
+++ b/pkg/instance/service.go
@@ -1,6 +1,7 @@
 package instance
 
 import (
+	"errors"
 	"fmt"
 	"io"
 	"log"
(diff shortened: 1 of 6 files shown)


tonsV2 (approved):
Good job! :+1:

PR #458, opened 2023-09-06
chore: abstract new stack implementation using dto struct in the handler

tonsV2:


pkg/stack/handler.go
```diff
@@ -56,7 +58,17 @@ func (h Handler) Find(c *gin.Context) {
 		return
 	}
 
-	c.JSON(http.StatusOK, stack)
+	// TODO: Remove this and just return the stack once the front end has caught up
+	s := Stack{}
```
teleivo, line 62:
```suggestion
	s := Stack{
	  Name: stack.Name,
	  Parameters: maps.Values(stack.Parameters)
	}
```

always prefer constructing the struct in one go if possible.

teleivo (approved):

PR #459, opened 2023-09-06
chore: remove stack parameter name

tonsV2:


pkg/model/stack.go
```diff
@@ -16,7 +16,6 @@ type Stack struct {
 type StackParameters map[string]StackParameter
 
 type StackParameter struct {
-	Name         string  `json:"name"`
```
teleivo, line 19: You will need to adapt the transformation of paramters to JSON in the handler. Otherwise the frontend will not have the `name` in the JSON.

teleivo (changes requested):

teleivo (approved):

PR #460, opened 2023-09-06
feat: deployments DEVOPS-223

tonsV2:


pkg/instance/router.go
```diff
@@ -23,4 +23,8 @@ func Routes(r *gin.Engine, authenticator gin.HandlerFunc, handler Handler) {
 	tokenAuthenticationRouter.PUT("/instances/:id/restart", handler.Restart)
 	tokenAuthenticationRouter.GET("/instances/:id/logs", handler.Logs)
 	tokenAuthenticationRouter.GET("/instances-name-to-id/:groupName/:instanceName", handler.NameToId)
+
+	tokenAuthenticationRouter.POST("/deployments", handler.SaveDeployment)
```
teleivo, line 27: I wonder if we should put deployments into its own handler 🤔 to clearly separate it from `instances` which it is.

pkg/model/instance.go
```diff
@@ -3,8 +3,74 @@ package model
...
+	Name        string `json:"name" gorm:"index:idx_name_and_group,unique"`
+	Description string `json:"description"`
+	GroupName   string `json:"groupName" gorm:"index:idx_name_and_group,unique; references:Name"`
+	Group       *Group `json:"group,omitempty"`
+	TTL         uint   `json:"ttl"`
+
+	Instances []*DeploymentInstance `json:"instances"`
+}
+
+type Parameters map[string]DeploymentInstanceParameter
```
teleivo, line 27: Since all our models are in the same package I would prefix this with `Deployment` or `DeploymentInstance`. To make it clear when using the package `model.StackParameter`, `model.DeploymentParameter`.

pkg/model/instance.go
```diff
@@ -3,8 +3,74 @@ package model
...
+
+	DeploymentID uint        `json:"deploymentId"`
+	Deployment   *Deployment `json:"deployment,omitempty"`
+
+	StackName string `json:"stackName" gorm:"references:Name"`
+
+	GormParameters []DeploymentInstanceParameter `json:"-" gorm:"foreignKey:DeploymentInstanceID; constraint:OnUpdate:CASCADE,OnDelete:CASCADE"`
+	Parameters     Parameters                    `json:"parameters" gorm:"-:all"`
+
+	Preset   bool `json:"preset"`   // Whether this instance is a preset
```
teleivo, line 42: Preset, PresetID, Public should be removed

pkg/model/instance.go
```diff
@@ -3,8 +3,74 @@ package model
...
+	UpdatedAt time.Time `json:"updatedAt"`
+
+	UserID uint  `json:"userId"`
+	User   *User `json:"user,omitempty"`
+
+	Name        string `json:"name" gorm:"index:idx_name_and_group,unique"`
+	Description string `json:"description"`
+	GroupName   string `json:"groupName" gorm:"index:idx_name_and_group,unique; references:Name"`
+	Group       *Group `json:"group,omitempty"`
+	TTL         uint   `json:"ttl"`
```
teleivo, line 22: TTL should be removed

pkg/model/instance.go
```diff
@@ -3,8 +3,74 @@ package model
...
+	DeploymentInstanceID uint   `json:"-" gorm:"primaryKey"`
+	ParameterName        string `json:"-" gorm:"primaryKey"`
+	StackName            string `json:"-"`
+	Value                string `json:"value"`
+}
+
+func (i *DeploymentInstance) BeforeSave(_ *gorm.DB) error {
+	i.GormParameters = make([]DeploymentInstanceParameter, 0, len(i.Parameters))
+	for _, parameter := range i.Parameters {
+		parameter.DeploymentInstanceID = i.ID
```
teleivo, line 59: Do we also need to set the `parameter.ParameterName`? or is it already set

pkg/storage/postgresql.go
```diff
@@ -20,7 +20,7 @@ func NewDatabase(c config.Postgresql) (*gorm.DB, error) {
 	dsn := fmt.Sprintf("host=%s user=%s password=%s dbname=%s port=%d sslmode=disable", host, username, password, name, port)
 
 	databaseConfig := gorm.Config{
-		Logger:         logger.Default.LogMode(logger.Warn),
+		Logger:         logger.Default.LogMode(logger.Info),
```
teleivo, line 23: I don't think we want Info in production.

scripts/instances/create-deployment.sh
```diff
@@ -0,0 +1,23 @@
...
+
+source ./auth.sh
+
+GROUP=$1
+NAME=$2
+shift
+shift
+DESCRIPTION=${*:-""}
+
+# container(s) in dhis2 pod will be restarted after that due to restartPolicy
```
teleivo, line 13: that comment does not apply here please remove

pkg/instance/handler.go
```diff
@@ -32,6 +32,206 @@ type groupServiceHandler interface {
 	Find(name string) (*model.Group, error)
 }
 
+type SaveDeploymentRequest struct {
+	Name        string `json:"name" binding:"required,dns_rfc1035_label"`
+	Description string `json:"description"`
+	Group       string `json:"group" binding:"required"`
+	TTL         uint   `json:"ttl"`
+	Source      uint   `json:"source"` // TODO: Create from source eg. other deployment
```
teleivo, line 40: TTL, Source, Preset should be removed

pkg/instance/handler.go
```diff
@@ -32,6 +32,206 @@ type groupServiceHandler interface {
...
+
+	deployment := &model.Deployment{
+		UserID:      user.ID,
+		Name:        request.Name,
+		Description: request.Description,
+		GroupName:   request.Group,
+		TTL:         request.TTL,
+	}
+
+	group, err = h.groupService.Find(request.Group)
```
teleivo, line 96: This code looks duplicated to me  🤔 

```go
	group, err := h.groupService.Find(request.Group)
	if err != nil {
		_ = c.Error(err)
		return
	}

	if !group.Deployable {
		forbidden := errdef.NewForbidden("group isn't deployable: %s", group.Name)
		_ = c.Error(forbidden)
		return
	}
```

is already done before

pkg/instance/handler.go
```diff
@@ -32,6 +32,206 @@ type groupServiceHandler interface {
...
+	//	403: Error
+	//	404: Error
+	//	415: Error
+	var request SaveDeploymentRequest
+	if err := handler.DataBinder(c, &request); err != nil {
+		_ = c.Error(err)
+		return
+	}
+
+	group, err := h.groupService.Find(request.Group)
```
teleivo, line 66: Should this already check if the user has access to the group?
I think we do check it in some way in `handler.CanWriteDeployment` 🤔 but not as explicitly

pkg/instance/handler.go
```diff
@@ -32,6 +32,206 @@ type groupServiceHandler interface {
...
+
+type parameter struct{ Value string }
+
+type parameters map[string]parameter
+
+type SaveInstanceRequest struct {
+	StackName string `json:"stackName"`
+	Preset    bool   `json:"preset"`
+	PresetID  uint   `json:"presetId"` // The preset id this link is created from
+	//	Public     bool                      `json:"public"`
```
teleivo, line 177: Preset, PresetID and this comment should be removed

pkg/instance/handler.go
```diff
@@ -32,6 +32,206 @@ type groupServiceHandler interface {
...
+		_ = c.Error(err)
+		return
+	}
+
+	deploymentId, ok := handler.GetPathParameter(c, "id")
+	if !ok {
+		return
+	}
+
+	// Convert request parameters to LinkParameters
```
teleivo, line 208: there are no LinkParameters

pkg/instance/handler.go
```diff
@@ -32,6 +32,206 @@ type groupServiceHandler interface {
...
+	params := make(model.Parameters, len(request.Parameters))
+	for k, v := range request.Parameters {
```
teleivo, line 210: I though you wanted us to be more explicit? Like `parameterName, parameter`

pkg/instance/handler.go
```diff
@@ -32,6 +32,206 @@ type groupServiceHandler interface {
...
+			Value:         v.Value,
+		}
+	}
+
+	instance := &model.DeploymentInstance{
+		DeploymentID: deploymentId,
+		StackName:    request.StackName,
+		Parameters:   params,
+		Preset:       request.Preset,
+		//		PresetID:   0,
```
teleivo, line 222: remove Preset and commented out code

pkg/instance/repository.go
```diff
@@ -21,6 +21,34 @@ type repository struct {
...
 
+func (r repository) SaveDeployment(deployment *model.Deployment) error {
+	// TODO: Do we need the option to save nested entities... Yes, if we create the deployment from a preset we need to store all the links as well
+	//err := r.db.Session(&gorm.Session{FullSaveAssociations: true}).Save(deployment).Error
+	err := r.db.Create(&deployment).Error
+	if errors.Is(err, gorm.ErrDuplicatedKey) {
+		return errdef.NewDuplicated("a deployment named %q already exists", deployment.Name)
+	}
+
+	return err
```
teleivo, line 32: Some time back we discussed that we would provide our own errors instead of just returning 3rd party errors. https://github.com/dhis2-sre/im-doc/blob/master/dev/README.md#errors

I would at least do `fmt.Errorf("failed to save deployment: %v", err)`. In the long run we should log details with a correlation id and only return a high level error (explanation) to the user. This applies to all repository methods.

pkg/instance/service.go
```diff
@@ -58,6 +61,60 @@ type service struct {
...
+
+func (s service) validateDeployment(deployment *model.Deployment) error {
+	stacks, err := s.collectStacks(deployment)
+	if err != nil {
+		return err
+	}
+
+	err = stack.ValidateNoCycles(stacks)
+	if err != nil {
+		return err
```
teleivo, line 95: since this goes to the user I would make it a little more friendly (while keeping the original error in as we don't have the log mechanism described in repository).

something lik

```suggestion
		return fmt.Errorf("error saving instance for %q as it adds a cycle which cannot be deployed: %v", instance.StackName, err)
```

pkg/instance/service.go
```diff
@@ -58,6 +61,60 @@ type service struct {
...
+	}
+
+	err = stack.ValidateConsumedParameters(stacks)
+	if err != nil {
+		return err
```
teleivo, line 100: at least

```suggestion
		return fmt.Errorf("error saving instance for %q as it is missing a required stack: %v", instance.StackName, err)
```

Ideally we would tell them which stack they are missing

teleivo (commented):
Please remove anything that we said we would discuss in more detail with Phil before implementing it: ttl, preset (source), public. This means code that is uncommented/commented. It might be that we will add it as you have it now but it might not.

Imagine that presets are implemented in a different way. You will have to deal with DB columns that you added here. This will only lead to extra work.

pkg/instance/router.go
tonsV2: Aren't they going to be closely related anyway?

pkg/instance/handler.go
```diff
@@ -32,6 +32,179 @@ type groupServiceHandler interface {
...
+	}
+
+	canWrite := handler.CanWriteDeployment(user, deployment)
+	if !canWrite {
+		unauthorized := errdef.NewUnauthorized("write access denied")
+		_ = c.Error(unauthorized)
+		return
+	}
+
+	// TODO: If request.Source, load deployment... Maybe only support this for instances
```
teleivo, line 95: please remove

pkg/instance/repository.go
```diff
@@ -21,6 +21,37 @@ type repository struct {
 	instanceParameterEncryptionKey string
 }
 
+func (r repository) SaveDeployment(deployment *model.Deployment) error {
+	// TODO: Do we need the option to save nested entities... Yes, if we create the deployment from a preset we need to store all the links as well
```
teleivo, line 25: please remove

pkg/instance/repository.go
```diff
@@ -21,6 +21,37 @@ type repository struct {
...
+	//err := r.db.Session(&gorm.Session{FullSaveAssociations: true}).Save(deployment).Error
+	err := r.db.Create(&deployment).Error
+	if err != nil {
+		if errors.Is(err, gorm.ErrDuplicatedKey) {
+			return errdef.NewDuplicated("a deployment named %q already exists", deployment.Name)
+		}
+		return fmt.Errorf("failed to save deployment: %v", err)
```
teleivo, line 32: the comment here also applied to all other repository methods 😅

teleivo (approved):

pkg/instance/repository.go
tonsV2: What comment?

PR #463, opened 2023-09-07
chore: make instance handler testable

teleivo:
These are changes we made to other parts of the code when we added an integration test.

* remove unused dependencies/fields from a handler, service, ...
* remove instance service interface in the instance handler as we do not benefit from it. We do not test the handler in isolation, nor do we have multiple instance service implementations. Its also not 3rd party code we want to isolate our code from. Only add an interface for any of
the above reasons.
* only pass in the needed config instead of the entire config struct

### DIFF
--- a/cmd/serve/main.go
+++ b/cmd/serve/main.go
@@ -104,9 +104,9 @@ func run() error {
 
 	stackService := stack.NewService(stacks)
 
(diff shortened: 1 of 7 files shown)


teleivo:
the check out in the action is done from an old commit :shru

PR #464, opened 2023-09-07
chore: make instance handler testable

teleivo:
* remove unused dependencies/fields
* remove instance service interface in the instance handler as we do not benefit from it. We do not test the handler in isolation, nor do we have multiple instance service implementations. Its also not 3rd party code we want to isolate our code from. Only add an interface for any of the above reasons.
* only pass in the needed config instead of the entire config struct

### DIFF
--- a/cmd/serve/main.go
+++ b/cmd/serve/main.go
@@ -104,9 +104,9 @@ func run() error {
 
 	stackService := stack.NewService(stacks)
 
-	instanceRepo := instance.NewRepository(db, cfg)
-	helmfileService := instance.NewHelmfileService(stackService, cfg)
-	instanceService := instance.NewService(cfg, instanceRepo, groupService, stackService, helmfileService)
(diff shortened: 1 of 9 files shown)


cmd/serve/main.go
```diff
@@ -104,9 +104,9 @@ func run() error {
 
 	stackService := stack.NewService(stacks)
 
-	instanceRepo := instance.NewRepository(db, cfg)
-	helmfileService := instance.NewHelmfileService(stackService, cfg)
-	instanceService := instance.NewService(cfg, instanceRepo, groupService, stackService, helmfileService)
+	instanceRepo := instance.NewRepository(db, cfg.InstanceParameterEncryptionKey)
```
tonsV2, line 107: s/Repo/Repository/ :smile:

tonsV2 (approved):

PR #467, opened 2023-09-08
chore: initial attempt at deploying

tonsV2:


pkg/instance/service.go
```diff
@@ -389,3 +392,64 @@ func (s service) destroy(instance *model.Instance) error {
...
+
+		for name, parameter := range instance.Parameters {
+			if stack.Parameters[name].Consumed {
+				for _, requiredStack := range stack.Requires {
+					// consume from provider
+					provider, ok := requiredStack.ParameterProviders[name]
+					if ok {
+						value, err := provider.Provide(*instance)
+						if err != nil {
+							return err
```
teleivo, line 433:
```suggestion
							return fmt.Errorf("failed to provide value for instance %q parameter %q: %v, instance.Name, name, err)
```

pkg/instance/service.go
```diff
@@ -389,3 +392,64 @@ func (s service) destroy(instance *model.Instance) error {
...
+						}
+						parameter.Value = value
+						continue
+					}
+
+					// consume from instance parameters
+					src := deployment.FindInstanceByStackName(requiredStack.Name)
+					if src == nil {
+						return errdef.NewNotFound("instance not found by stackName name %q", stackName)
```
teleivo, line 442:
```suggestion
						return errdef.NewNotFound("failed to find required instance %q of instance %q", src.Name, instance.Name)
```

teleivo (commented):
I added some comments as I was reviewing the entire PR. I know its a draft, just so we don't forget.

pkg/instance/service.go
```diff
@@ -389,3 +393,134 @@ func (s service) destroy(instance *model.Instance) error {
...
+
+		// Add parameters not supplied by user
+		for name, stackParameter := range stack.Parameters {
+			if _, ok := instanceParameters[name]; !ok {
+				instanceParameter := model.DeploymentInstanceParameter{
+					ParameterName: name,
+					StackName:     stack.Name,
+				}
+
+				if stackParameter.DefaultValue != nil {
```
teleivo, line 434: We should return an error if `!stack.Parameters[name].Consumed`, there is no default and the user did not provide a value, right?

We need to be 100% sure that we have all parameters for each instance. If we don't we should not call `deploy` but return an error.

pkg/instance/helmfile.go
```diff
@@ -100,6 +129,55 @@ func (h helmfileService) loadStackParameters(stackName string) (stackParameters,
...
+	cmd.Env = injectEnv(cmd.Env, "KUBERNETES_SERVICE_PORT")
+	cmd.Env = injectEnv(cmd.Env, "KUBERNETES_PORT")
+	cmd.Env = injectEnv(cmd.Env, "KUBERNETES_PORT_443_TCP_ADDR")
+	cmd.Env = injectEnv(cmd.Env, "KUBERNETES_PORT_443_TCP_PORT")
+	cmd.Env = injectEnv(cmd.Env, "KUBERNETES_PORT_443_TCP_PROTO")
+	cmd.Env = injectEnv(cmd.Env, "KUBERNETES_PORT_443_TCP")
+	cmd.Env = injectEnv(cmd.Env, "KUBERNETES_SERVICE_PORT_HTTPS")
+	cmd.Env = injectEnv(cmd.Env, "KUBERNETES_SERVICE_HOST")
+
+	for name, parameter := range stack.Parameters {
```
teleivo, line 163: Why are we not ranging over `instance.Parameters`? I thought in here the `DeploymentInstance` is all that is needed 🤔

pkg/instance/handler.go
```diff
@@ -38,6 +38,50 @@ type SaveDeploymentRequest struct {
...
+	//	401: Error
+	//	403: Error
+	//	404: Error
+	//	415: Error
+	id, ok := handler.GetPathParameter(c, "id")
+	if !ok {
+		return
+	}
+
+	deployment, err := h.instanceService.FindDeploymentById(id)
```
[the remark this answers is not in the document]
tonsV2: The order here doesn't really matter because if the token isn't valid the middleware won't even execute the handler in the first place

pkg/instance/service.go
```diff
@@ -415,22 +499,18 @@ func (s service) DeployDeployment(token string, deployment *model.Deployment) er
...
-				ParameterName: name,
-				StackName:     stack.Name,
-			}
+				if stackParameter.DefaultValue != nil {
+					instanceParameter.Value = *stackParameter.DefaultValue
+				}
 
-			if stackParameter.DefaultValue != nil {
-				instanceParameter.Value = *stackParameter.DefaultValue
+				instanceParameters[name] = instanceParameter
```
teleivo, line 512: Why do we set the `instanceParameters[name]` if the `stackParameter.DefaultValue == nil`?

And why is this loop done for all and not just `!Consumed` parameters? 🤔

pkg/instance/service.go
```diff
@@ -80,6 +80,12 @@ func (s service) SaveInstance(instance *model.DeploymentInstance) error {
 	}
 
 	deployment.Instances = append(deployment.Instances, instance)
+
+	err = s.resolveParameters(deployment)
```
teleivo, line 84: I think its best to check `	_, err := s.validateNoCycles(deployment.Instances)` before resolving parameters like we do in `DeployDeployment`. Otherwise you might get a weird behavior in case there are cycles, missing stacks.

pkg/instance/service.go
```diff
@@ -94,29 +100,89 @@ func (s service) validateDeployment(deployment *model.Deployment) error {
...
+	var errs []error
+	for _, instance := range deployment.Instances {
 		stack, err := s.stackService.Find(instance.StackName)
+		if err != nil {
+			errs = append(errs, fmt.Errorf("stack (%s) not found for instance: %s", instance.StackName, instance.Name))
+		}
+
+		for name, stackParameter := range stack.Parameters {
+			var found bool
+			parameterName := name
```
teleivo, line 121: If this is needed declare it closer to where its used.

pkg/instance/service.go
```diff
@@ -94,29 +100,89 @@ func (s service) validateDeployment(deployment *model.Deployment) error {
...
 		return err
 	}
 
 	return nil
 }
 
-func (s service) collectStacks(deployment *model.Deployment) ([]model.Stack, error) {
-	stacks := make([]model.Stack, len(deployment.Instances))
-	for i, instance := range deployment.Instances {
+func (s service) validateDeploymentParameters(deployment *model.Deployment) error {
```
teleivo, line 111: `validateDeploymentParameters` is and should be called after `resolveParameters`.

I think the validation can be simpler 🤔. The rule is every stack parameter needs to have a value. Since you already resolved default values there is no need to add any logic regarding that. There is also no need to go through graph of where a parameter is consumed from for the same reason.

The only question is if a parameter does not have a value: who's fault is it?
* if the parameter is a consumed one it is likely our fault
* if the parameter is not a consumed one it is likely the users fault

So these two cases need to influence the error message. Unless I miss something.

I would also call `validateDeploymentParameters` on deploy. This way we can prevent some calls to `helmfile` which would fail due to some changes in the stacks.

I am unsure if we should call `resolveParameters` again on deploy. We already did that on save and stored the parameters right? 🤔

teleivo (commented):
Nice tests 🥳

Can you please group the new `DeploymentInstance` related methods together. Also order methods/functions in call order like `validateNoCycles` and `validateDeploymentParameters` which makes it easy to read from top to bottom. The code is harder to read because I have to jump around 😅 

There are some things in the algorithm that I am not sure are correct or as simple as they could be (see individual comments). Adding to the high level description of the algorithm. I think for save and deploy we want in that order

- validate the deployment has no cycles (which includes the check that all required stacks are present, stacks are only present once)
- resolve instance parameters (set value to default for !Consumed parameters and set `Consumed=true` directly from source or via source provider)
- validate all instances have a value for every parameter of their stack (irrespective of where the parameter came from. `Consumed=false` is likely the users fault, `Consumed=true` is likely our fault)

on deployment we should do the above and
- order the deployment instances topologically
- deploy instances in order

pkg/instance/service.go
```diff
@@ -76,43 +80,267 @@ func (s service) SaveInstance(instance *model.DeploymentInstance) error {
...
 
-func (s service) validateDeployment(deployment *model.Deployment) error {
-	stacks, err := s.collectStacks(deployment)
+func (s service) validateNoCycles(instances []*model.DeploymentInstance) (graph.Graph[string, *model.DeploymentInstance], error) {
+	g := graph.New(func(instance *model.DeploymentInstance) string {
+		return instance.StackName
+	}, graph.Directed(), graph.PreventCycles())
+
+	for _, instance := range instances {
+		err := g.AddVertex(instance)
```
teleivo, line 108: If `errors.Is(err, graph.ErrVertexAlreadyExists)` we should return something like `fmt.Errorf("failed adding instance for stack %q as one already exists.", instance.StackName)`.

pkg/instance/service.go
```diff
@@ -76,43 +80,267 @@ func (s service) SaveInstance(instance *model.DeploymentInstance) error {
...
+		if err != nil {
+			return nil, err
+		}
+		for _, requiredStack := range stack.Requires {
+			err := g.AddEdge(src.StackName, requiredStack.Name)
+			if err != nil {
+				if errors.Is(err, graph.ErrEdgeAlreadyExists) {
+					return nil, fmt.Errorf("instance %q requires %q more than once", src.Name, requiredStack.Name)
+				} else if errors.Is(err, graph.ErrEdgeCreatesCycle) {
+					return nil, fmt.Errorf("edge from instance %q to instance %q creates a cycle", src.Name, requiredStack.Name)
```
teleivo, line 125: Should we use the work `link` when talking to users instead of `edge`? Since we already established the word link.

```suggestion
					return nil, fmt.Errorf("link from instance %q to instance %q creates a cycle", src.Name, requiredStack.Name)
```

Also we are using `instance` in the message but some of the args are stack names.

pkg/instance/service.go
```diff
@@ -76,43 +80,267 @@ func (s service) SaveInstance(instance *model.DeploymentInstance) error {
...
+				}
+				return nil, fmt.Errorf("failed adding edge from instance %q to instance %q: %v", src.Name, requiredStack.Name, err)
```
teleivo, line 127: same comments as above edge vs link/mix of instance and stack names

```suggestion
				return nil, fmt.Errorf("failed linking instance %q with instance %q: %v", src.Name, requiredStack.Name, err)
```

pkg/instance/service.go
```diff
@@ -76,43 +80,267 @@ func (s service) SaveInstance(instance *model.DeploymentInstance) error {
...
+			}
+		}
+	}
+
+	return g, nil
+}
+
+func (s service) resolveParameters(deployment *model.Deployment) error {
+	for _, instance := range deployment.Instances {
+		instanceParameters := instance.Parameters
```
teleivo, line 137: move declaration close to where its used

pkg/instance/service.go
```diff
@@ -76,43 +80,267 @@ func (s service) SaveInstance(instance *model.DeploymentInstance) error {
...
+		}
+	}
+
+	return nil
+}
+
+func rejectConsumedParameters(instanceParameters model.DeploymentInstanceParameters, stack *model.Stack) error {
+	var errs []error
+	for name := range instanceParameters {
+		if stack.Parameters[name].Consumed {
```
teleivo, line 192: Would this not panic given an instance parameter that is not part of the stack.Parameters 🤔 (since `rejectConsumedParameters` is called before `rejectNonExistingParameters`)

pkg/instance/service.go
```diff
@@ -76,43 +80,267 @@ func (s service) SaveInstance(instance *model.DeploymentInstance) error {
...
+	var errs []error
+	for name := range instanceParameters {
+		if _, ok := stack.Parameters[name]; !ok {
+			errs = append(errs, fmt.Errorf("parameter not found on stack: %s", name))
+		}
+	}
+	return errors.Join(errs...)
+}
+
+func addUnsuppliedParameters(stack *model.Stack, instanceParameters model.DeploymentInstanceParameters) {
```
teleivo, line 209: I would make this only about default values for `!Consumed` parameters => `addDefaultParameterValues`.

I find the interplay between `addUnsuppliedParameters` and resolving consumed parameters above odd/confusing 🤔 By that I mean that `instanceParameters[name] = instanceParameter`.

teleivo (commented):
Prefer using `%q` instead of `%s` for strings args in error messages. Especially if they are user supplied so we would see any odd whitespace issues.

pkg/instance/service.go
```diff
@@ -76,43 +80,224 @@ func (s service) SaveInstance(instance *model.DeploymentInstance) error {
...
+
+		err = rejectConsumedParameters(instanceParameters, stack)
+		if err != nil {
+			return err
+		}
+
+		addDefaultParameterValues(stack, instanceParameters)
+
+		for name, parameter := range instanceParameters {
+			stackParameter := stack.Parameters[name]
```
teleivo, line 154: Can you please move this block into 2 methods `validateParameters`, `resolveConsumedParameters`.

This method is then super readable 😄 

```go
instanceParameters := instance.Parameters
err = rejectNonExistingParameters(instanceParameters, stack)
if err != nil {
	return err
}

err = rejectConsumedParameters(instanceParameters, stack)
if err != nil {
	return err
}

addDefaultParameterValues(stack, instanceParameters)

validateParameters(instanceParameters, stack)

resolveConsumedParameters(instanceParameters, stack)
```

pkg/instance/service.go
```diff
@@ -76,43 +80,224 @@ func (s service) SaveInstance(instance *model.DeploymentInstance) error {
...
+		if err != nil {
+			return err
+		}
+
+		err = rejectConsumedParameters(instanceParameters, stack)
+		if err != nil {
+			return err
+		}
+
+		addDefaultParameterValues(stack, instanceParameters)
```
teleivo, line 151: Please switch the parameters so we have a consistent order of `instanceParameters, stack` in all the methods we call here.

pkg/instance/service_test.go
```diff
@@ -2,16 +2,206 @@ package instance
...
+		stacks := stack.Stacks{
+			"stack-a": stackA,
+			"stack-b": stackB,
+		}
+		stackService := stack.NewService(stacks)
+		service := NewService(nil, nil, stackService, nil)
+		deployment := &model.Deployment{
+			Instances: []*model.DeploymentInstance{
+				{
+					Name:       "name",
```
teleivo, line 169: you could make them `name-a`, `group-a` just to make it more clear in the expected parameter that the values come from an instance of stack a.

PR #474, opened 2023-09-20
fix: linking stopped working with the new instance deployment strateg…

tonsV2:
…y. This "hack" fixes it

### DIFF
--- a/pkg/instance/service.go
+++ b/pkg/instance/service.go
@@ -374,11 +374,16 @@ func (s service) findParameterValue(parameter string, sourceInstance *model.Inst
 	}
 
 	stackParameter, ok := sourceStack.Parameters[parameter]
-	if !ok {
-		return "", fmt.Errorf("unable to find value for parameter: %s", parameter)
+	if ok {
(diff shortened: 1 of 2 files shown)


radnov (approved):
LGTM 👍

pkg/instance/service.go
```diff
@@ -374,11 +374,16 @@ func (s service) findParameterValue(parameter string, sourceInstance *model.Inst
...
 	stackParameter, ok := sourceStack.Parameters[parameter]
-	if !ok {
-		return "", fmt.Errorf("unable to find value for parameter: %s", parameter)
+	if ok {
+		return *stackParameter.DefaultValue, nil
 	}
 
-	return *stackParameter.DefaultValue, nil
+	// TODO: Remove HostnamePattern once stacks 2.0 are the default
+	if parameter == "DATABASE_HOSTNAME" {
```
teleivo, line 382: Since you have the stack parameter you should be able to call the provider function without having to add the field HostnamePattern to the stack. I think 😋

PR #478, opened 2023-09-21
feat: destroy deployment

tonsV2:
This PR can be tested using the following commands

    ./create-deployment.sh whoami my-deployment-1 some test deployment
    # Update the below id to match the output of the above command
    ./create-deployment-instance-dhis2-db.sh 1
    ./create-deployment-instance-dhis2-core.sh 1
    ./findDeploymentById.sh 1
    ./destroy-deployment.sh 1

### DIFF
--- a/pkg/instance/handler.go
+++ b/pkg/instance/handler.go
@@ -1132,7 +1132,7 @@ func (h Handler) DeleteDeployment(c *gin.Context) {
 
 	err = h.instanceService.DeleteDeployment(deployment.ID)
 	if err != nil {
-		_ = c.Error(fmt.Errorf("unable to delete instance: %v", err))
+		_ = c.Error(fmt.Errorf("unable to delete deployment: %v", err))
 		return
 	}
 
--- a/pkg/instance/handler.go
+++ b/pkg/instance/handler.go
@@ -1088,3 +1088,53 @@ func (h Handler) ListPublicInstances(c *gin.Context) {
 
 	c.JSON(http.StatusOK, instances)
 }
+
+// DeleteDeployment deployment by id
+func (h Handler) DeleteDeployment(c *gin.Context) {
+	// swagger:route DELETE /deployment/{id} deleteDeployment
+	//
+	// Delete deployment
+	//
+	// Delete an deployment by id
+	//
+	// Security:
+	//	oauth2:
+	//
+	// responses:
+	//	202:
+	//	401: Error
+	//	403: Error
+	//	404: Error
+	//	415: Error
+	id, ok := handler.GetPathParameter(c, "id")
+	if !ok {
+		return
+	}
+
+	user, err := handler.GetUserFromContext(c)
+	if err != nil {
+		_ = c.Error(err)
+		return
+	}
+
+	deployment, err := h.instanceService.FindDeploymentById(id)
+	if err != nil {
+		_ = c.Error(err)
+		return
+	}
+
+	canWrite := handler.CanWriteDeployment(user, deployment)
+	if !canWrite {
+		unauthorized := errdef.NewUnauthorized("write access denied")
+		_ = c.Error(unauthorized)
+		return
+	}
+
+	err = h.instanceService.DeleteDeployment(deployment.ID)
+	if err != nil {
+		_ = c.Error(fmt.Errorf("unable to delete instance: %v", err))
+		return
+	}
+
+	c.Status(http.StatusAccepted)
+}
--- a/pkg/instance/handler.go
+++ b/pkg/instance/handler.go
@@ -1130,7 +1130,7 @@ func (h Handler) DeleteDeployment(c *gin.Context) {
 		return
 	}
 
-	err = h.instanceService.DeleteDeployment(deployment.ID)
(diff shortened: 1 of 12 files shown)


pkg/instance/repository.go
```diff
@@ -21,6 +23,34 @@ type repository struct {
...
+		if errors.Is(err, gorm.ErrRecordNotFound) {
+			return errdef.NewNotFound("deployment not found by id: %d", id)
+		}
+		return fmt.Errorf("failed to delete deployment: %v", err)
+	}
+
+	for _, instance := range deployment.Instances {
+		err := r.db.Unscoped().Delete(&model.DeploymentInstance{}, instance.ID).Error
+		if err != nil {
+			return fmt.Errorf("failed to delete instance: %v", err)
```
teleivo, line 38:
```suggestion
			return fmt.Errorf("failed to delete instance %q: %v", instance.Name, err)
```

pkg/instance/repository.go
```diff
@@ -21,6 +23,34 @@ type repository struct {
...
+		}
+	}
+
+	// TODO: The below is suppose to delete the instances as well but I can't make it work. So the above is needed...
```
teleivo, line 42: Please remove this if it does not work.

pkg/instance/service.go
```diff
@@ -267,6 +269,7 @@ func (s service) DeployDeployment(token string, deployment *model.Deployment) er
 		return err
 	}
 
+	// TODO: Delete?
```
teleivo, line 272: Why did you add this TODO?

pkg/instance/service.go
```diff
@@ -313,6 +316,71 @@ func (s service) deployDeploymentInstance(token string, instance *model.Deployme
 	return nil
 }
 
+func (s service) DeleteDeployment(id uint) error {
+	deployment, err := s.FindDeploymentById(id)
```
teleivo, line 320: We find the deployment 3 times if I am counting correctly 😅 Since we call `s.instanceRepository.DeleteDeployment(id)` only from here, can we pass in the `Deployment` to not fetch it again?

pkg/instance/service.go
```diff
@@ -313,6 +316,71 @@ func (s service) deployDeploymentInstance(token string, instance *model.Deployme
...
+	deploymentGraph, err := s.validateNoCycles(deployment.Instances)
+	if err != nil {
+		return err
+	}
+
+	instances, err := deploymentOrder(deployment, deploymentGraph)
+	if err != nil {
+		return err
+	}
+
```
teleivo, line 334: Please remove the newline as `deploymentOrder` and `Reverse` are closely related. Or best replace it with a comment of why you are calling `Reverse`. This might not be obvious at first.

pkg/instance/service.go
```diff
@@ -313,6 +316,71 @@ func (s service) deployDeploymentInstance(token string, instance *model.Deployme
...
+	if errs != nil {
+		return errs
+	}
+
+	return s.instanceRepository.DeleteDeployment(id)
+}
+
+func (s service) destroyDeploymentInstance(instance *model.DeploymentInstance) error {
+	if instance.DeployLog != "" {
```
teleivo, line 353: please invert this condition to remove nesting.

pkg/instance/handler.go
```diff
@@ -1088,3 +1088,53 @@ func (h Handler) ListPublicInstances(c *gin.Context) {
...
+		return
+	}
+
+	user, err := handler.GetUserFromContext(c)
+	if err != nil {
+		_ = c.Error(err)
+		return
+	}
+
+	deployment, err := h.instanceService.FindDeploymentById(id)
```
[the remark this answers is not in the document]
tonsV2: yes

pkg/instance/repository.go
```diff
@@ -21,6 +21,25 @@ type repository struct {
 	instanceParameterEncryptionKey string
 }
 
+func (r repository) DeleteDeployment(deployment *model.Deployment) error {
+	for _, instance := range deployment.Instances {
+		err := r.db.Unscoped().Delete(&model.DeploymentInstance{}, instance.ID).Error
```
teleivo, line 26: Can this also return a `gorm.ErrRecordNotFound` I see you are checking for it below

pkg/instance/service.go
```diff
@@ -313,6 +315,63 @@ func (s service) deployDeploymentInstance(token string, instance *model.Deployme
...
+
+	var errs error
+	for _, instance := range instances {
+		err := s.destroyDeploymentInstance(instance)
+		if err != nil {
+			errs = errors.Join(errs, fmt.Errorf("failed to destroy instance(%s) %q: %v", instance.StackName, instance.Name, err))
+		}
+	}
+
+	if errs != nil {
```
teleivo, line 338: What if we succeeded in destroying only some instances, should we delete these from the DB? 🤔 Imagine that case and a user retrying, would destroying the same instances that were destroyed before work again? Hence can a user retry and actually get all instances destroyed and removed from the DB in that case.

teleivo (approved):

PR #479, opened 2023-09-21
feat: destroy deployment instance

tonsV2:
This PR needs... https://github.com/dhis2-sre/im-manager/pull/478

This PR can be tested using the following commands

    ./create-deployment.sh whoami my-deployment-1 some test deployment
    # Update the below id to match the output of the above command
    ./create-deployment-instance-dhis2-db.sh 1
    ./create-deployment-instance-dhis2-core.sh 1
    ./findDeploymentById.sh 1
    ./delete-deployment-instance.sh 1 <id-of-the-dhis2-db-instance> # This should result in a bad request
    ./delete-deployment-instance.sh 1 <id-of-the-dhis2-core-instance>
    ./findDeploymentById.sh 1

The final result should be a deployment without the core instance.


pkg/instance/docs.go
```diff
@@ -64,6 +64,16 @@ type _ struct {
...
 }
 
+// swagger:parameters deleteDeploymentInstance
+type _ struct {
+	// in: path
+	// required: true
+	ID uint `json:"id"`
+	// in: path
+	// required: true
+	InstanceID uint `json:"instanceId"`
```
teleivo, line 74: At some point we were moving to accepting names instead of ids. Does this also make sense here? Users deleting an instance via its name?

pkg/instance/service.go
```diff
@@ -97,6 +97,38 @@ func (s service) SaveInstance(instance *model.DeploymentInstance) error {
 	return s.instanceRepository.SaveInstance(instance)
 }
 
+func (s service) DeleteInstance(deploymentId, instanceId uint) error {
+	deployment, err := s.FindDeploymentById(deploymentId)
```
teleivo, line 101: It would be interesting to think about the ACL. We often repeat what we already did in the handler in the service due to us doing the ACL check in the handler. How would it look if we were to move this in the service? Or how would it look if we were to pass in the entire model in the service? Something we should think about IMHO.

pkg/instance/service.go
```diff
@@ -97,6 +97,38 @@ func (s service) SaveInstance(instance *model.DeploymentInstance) error {
...
+		return err
+	}
+
+	index := slices.IndexFunc(deployment.Instances, func(instance *model.DeploymentInstance) bool {
+		return instanceId == instance.ID
+	})
+	if index == -1 {
+		return errdef.NewNotFound("instance not found on deployment(%d) by id: %d", deployment.ID, instanceId)
+	}
+
```
teleivo, line 112: The next line belongs to the above `slices.IndexFunc` and find the instance logic. I would remove the newline to indicate that.

pkg/instance/service.go
```diff
@@ -124,6 +156,8 @@ func (s service) validateNoCycles(instances []*model.DeploymentInstance) (graph.
 					return nil, fmt.Errorf("instance %q requires %q more than once", src.Name, requiredStack.Name)
 				} else if errors.Is(err, graph.ErrEdgeCreatesCycle) {
 					return nil, fmt.Errorf("link from instance %q to stack %q creates a cycle", src.Name, requiredStack.Name)
+				} else if errors.Is(err, graph.ErrVertexNotFound) {
+					return nil, fmt.Errorf("can't remove %q since at least one other instance depends on it: %s", requiredStack.Name, src.StackName)
```
teleivo, line 160: That error message only makes sense for the feature you are adding. You should phrase it in a way that makes sense for all features using this function.

pkg/instance/service.go
```diff
@@ -97,6 +97,38 @@ func (s service) SaveInstance(instance *model.DeploymentInstance) error {
...
+	deployment, err := s.FindDeploymentById(deploymentId)
+	if err != nil {
+		return err
+	}
+
+	index := slices.IndexFunc(deployment.Instances, func(instance *model.DeploymentInstance) bool {
+		return instanceId == instance.ID
+	})
+	if index == -1 {
+		return errdef.NewNotFound("instance not found on deployment(%d) by id: %d", deployment.ID, instanceId)
```
teleivo, line 110: The notation `deployment(%d)` reminds me of a function call 😅 I would send a rendered error message to Phil/Hendrik/... and ask them for what they find readable

Maybe

```suggestion
		return errdef.NewNotFound("instance %d not found in deployment %d", instanceId, deployment.ID)
```

pkg/instance/docs.go
tonsV2: Yeah, but that was due to an endpoint translating the name to its id. Identifying stacks and groups by name brought some issues in regards to how Gorm is used... In other words, let's keep this simple for now :smile:

pkg/instance/service.go
tonsV2: We have addressed this several times. I'm not against but I feel it's beyond the scope of this PR.

teleivo: True not in this PR but we have never really addressed it and keep on writing more code than might be needed.

pkg/instance/service.go
```diff
@@ -97,6 +97,37 @@ func (s service) SaveInstance(instance *model.DeploymentInstance) error {
...
+	})
+
+	_, err = s.validateNoCycles(deployment.Instances)
+	if err != nil {
+		return errdef.NewBadRequest("failed deleting instance: %v", err)
+	}
+
+	err = s.destroyDeploymentInstance(instance)
+	if err != nil {
+		return err
```
teleivo, line 125: Did you check the errors we get from `s.destroyDeploymentInstance` and below? I mostly see `return err` being returned.

Users should not get a k8s error (at least not as the start of the error message). So either `s.destroyDeploymentInstance` makes sure to return friendly errors or we `return fmt.Errorf("failed to destroy instance %d in deployment %d: %v", instanceId, deployment.ID,  err)` here.

pkg/instance/service.go
```diff
@@ -124,6 +155,8 @@ func (s service) validateNoCycles(instances []*model.DeploymentInstance) (graph.
 					return nil, fmt.Errorf("instance %q requires %q more than once", src.Name, requiredStack.Name)
 				} else if errors.Is(err, graph.ErrEdgeCreatesCycle) {
 					return nil, fmt.Errorf("link from instance %q to stack %q creates a cycle", src.Name, requiredStack.Name)
+				} else if errors.Is(err, graph.ErrVertexNotFound) {
+					return nil, fmt.Errorf("link from %q to stack %q required", requiredStack.Name, src.StackName)
```
teleivo, line 159: You are calling `validateNoCycles` from 4 methods. Does this error message make sense for all these cases? 🤔 `graph.ErrVertexNotFound` generally means that the required stack (vertex) is missing. The link (edge) is actually present but only one of its vertices is.

1. `SaveInstance(instance *model.DeploymentInstance)` - will this error be returned if I am saving a `core` without having saved a `db` first? Or will that be caught in https://github.com/dhis2-sre/im-manager/blob/a546c81c142bc767b26b872d391e88a1d959388d/pkg/instance/service.go#L228 `resolveConsumedParameters`
2. `DeployDeployment(token string, deployment *model.Deployment)`
3. `DeleteDeployment(deployment *model.Deployment)`
4. `DeleteInstance(deploymentId, instanceId uint)`

pkg/instance/service.go
```diff
@@ -97,6 +97,37 @@ func (s service) SaveInstance(instance *model.DeploymentInstance) error {
...
+	}
+	instance := deployment.Instances[index]
+
+	deployment.Instances = slices.DeleteFunc(deployment.Instances, func(instance *model.DeploymentInstance) bool {
+		return instanceId == instance.ID
+	})
+
+	_, err = s.validateNoCycles(deployment.Instances)
+	if err != nil {
+		return errdef.NewBadRequest("failed deleting instance: %v", err)
```
radnov, line 120: We're using `failed to ...` in other places, so I'd change this for consistency. Also, is this the correct error type? Seems a bit weird since the error is coming from `validateNoCycles()` - maybe there's a better option? 🤔

radnov (commented):
Just one minor comment.

And I did a quick test as described in the PR and got the expected results. 👍

pkg/instance/service.go
tonsV2: I've updated the wording for the sake of consistency.

I'd say it is, because the user is requesting an update which will leave the deployment in an invalid state. Thus it's a bad request.

PR #484, opened 2023-09-25
Test deploy dhis2

tonsV2:
1. Fixes the false positive when asserting if a pod is running
2. Ensures an error is return if a user attempts to deploy a deployment without instances
3. Test a deployment with a WhoAmI instance can be deployed

Once the cluster is up and running a Kubernetes configuration file can be retrieved using the below command

    docker cp (docker ps | grep rancher | awk '{print $NF}'):/var/gnomock/kubeconfig.yaml . && export KUBECONFIG=./kubeconfig.yaml

By setting break points in the code it can be asserted that the containers are actually in the required state


pkg/instance/instance_integration_test.go
```diff
@@ -61,68 +71,143 @@ func TestInstanceHandler(t *testing.T) {
...
+	require.NoError(t, err, "failed to create S3 output bucket")
+	s3 := inttest.SetupS3(t, s3Dir)
+	uploader := manager.NewUploader(s3.Client)
+	s3Client := storage.NewS3Client(s3.Client, uploader)
+	databaseRepository := database.NewRepository(db)
+	databaseService := database.NewService(s3Bucket, s3Client, groupService, databaseRepository)
+
+	authenticator := func(ctx *gin.Context) {
+		ctx.Set("user", &model.User{
+			ID:    user.ID, // TODO: Why is this necessary all of a sudden?
```
teleivo, line 93: What fails if you do not set it?

pkg/instance/instance_integration_test.go
```diff
@@ -61,68 +71,143 @@ func TestInstanceHandler(t *testing.T) {
...
+					Name: "group-name",
+				},
+			},
+		})
+	}
 	client := inttest.SetupHTTPServer(t, func(engine *gin.Engine) {
 		var twoDayTTL uint = 172800
 		instanceHandler := instance.NewHandler(groupService, instanceService, twoDayTTL)
-		instance.Routes(engine, func(ctx *gin.Context) {
-			ctx.Set("user", user)
```
teleivo, line 74: Why did you replace that with the `authenticator` function from above?

pkg/instance/instance_integration_test.go
```diff
@@ -61,68 +71,143 @@ func TestInstanceHandler(t *testing.T) {
...
+	var databaseID string
+	{
+		t.Log("Upload")
+		var b bytes.Buffer
+		w := multipart.NewWriter(&b)
+		err := w.WriteField("group", "group-name")
+		require.NoError(t, err, "failed to write form field")
+		f, err := w.CreateFormFile("database", "mydb")
+		require.NoError(t, err, "failed to create form file")
+		// TODO: Why does this work?
```
teleivo, line 131: What do you mean? https://pkg.go.dev/mime/multipart#Writer.CreateFormFile and https://pkg.go.dev/mime/multipart#Writer.CreatePart should explain how it works

pkg/instance/instance_integration_test.go
```diff
@@ -61,68 +71,143 @@ func TestInstanceHandler(t *testing.T) {
...
+			    }
+			]
+		}`), &instance, inttest.WithAuthToken("sometoken"))
+
+		assertPodIsReady(t, k8sClient, group.Name, instance.Name+"-database")
+		assertPodIsReady(t, k8sClient, group.Name, instance.Name)
+	})
+}
+
+func assertPodIsReady(t *testing.T, k8sClient *inttest.K8sClient, group string, instance string) {
```
teleivo, line 171: If we are repeatedly using our inttest client to do the same thing I would make this a method on the inttest client

```go
k8sClient := inttest.SetupK8s(t)
...
k8sClient.WaitUntilPodIsReady(t, group, instance)
```

If a test is doing something very specific they can then opt-out of our inttest client and use the underlying one like k8sClient.Client (or S3, ...).

teleivo (commented):
Makes sense overall 😄 

The only thing that bothers me is that we repeat/blur the lines between testing `instance` (handler) and `database` (handler). You repeated the same upload test we have in `database_integration_test.go`. I am not sure exactly what the way around it is. This might be a legitimate case of us wanting to use mocks to isolate the features/slices (instance, database, ...) from each other 🤔 Gnomock for example allows passing some initial fixtures into the setup like https://pkg.go.dev/github.com/orlangure/gnomock/preset/localstack#WithS3Files Maybe we could use that together with mocking. This might require us to work on our code structure/dependencies between packages.

pkg/instance/instance_integration_test.go
tonsV2: Error #01: ERROR: insert or update on table "instances" violates foreign key constraint "fk_instances_user" (SQLSTATE 23503)

That's what I'm getting locally for both tests

pkg/instance/instance_integration_test.go
tonsV2: It's the same and I use it in two places now

pkg/instance/instance_integration_test.go
tonsV2: I'm surprised the seed script doesn't return an error when we're seeding a file with the content "file contents"

pkg/instance/instance_integration_test.go
tonsV2: Currently we're only using it in one test. It'll probably be used in some of the database tests as well but since that's not the case now I just placed in the instance test.

Despite the above I've been thinking if we should move the save, and save as, functionality into the instance handler so we have endpoints such as /instance/:id/save rather than /save/:instanceId on the database handler but that comes with some issues which I wanted to discuss with you.

pkg/instance/instance_integration_test.go
tonsV2: But to expedite things here, let's just add it

pkg/instance/instance_integration_test.go
teleivo: I want to understand why you had to change it? Because I think you don't have to deal with https://github.com/dhis2-sre/im-manager/pull/484#discussion_r1335490921 if you don't replace this function as the `user` we have already has an ID.

pkg/instance/instance_integration_test.go
teleivo: Ah ok, well then you need to investigate 😋

pkg/instance/instance_integration_test.go
teleivo: > Currently we're only using it in one test.
You are using the function 3 times which to me is good enough to move this to the inttest client 😄

tonsV2: Sure! Should we make it a member of the K8sClient?

teleivo (approved):

pkg/inttest/k8s.go
```diff
@@ -46,3 +52,43 @@ type K8sClient struct {
 	Client *kubernetes.Clientset
 	Config []byte
 }
+
+func (k K8sClient) AssertPodIsReady(t *testing.T, group string, instance string) {
```
teleivo, line 56: Please add a little godoc maybe stating that it times out after 20 seconds.

tonsV2: I was thinking about passing the timeout as a parameter since it might differ depending on the instance we're deploying

pkg/instance/instance_integration_test.go
```diff
@@ -61,67 +63,102 @@ func TestInstanceHandler(t *testing.T) {
...
-						cancel()
-						return
-					}
-				}
-			}
-		}
+		k8sClient.AssertPodIsReady(t, group.Name, instance.Name)
+	})
+
+	t.Run("DeploymentWithoutInstances", func(t *testing.T) {
```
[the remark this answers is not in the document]
tonsV2: Sure!

pkg/instance/instance_integration_test.go
```diff
@@ -61,67 +63,102 @@ func TestInstanceHandler(t *testing.T) {
...
+		assert.Equal(t, "some description", deployment.Description)
+
+		t.Log("Deploy deployment")
+		path := fmt.Sprintf("/deployments/%d/deploy", deployment.ID)
+		response := client.Do(t, http.MethodPost, path, nil, http.StatusBadRequest, inttest.WithAuthToken("sometoken"))
+
+		assert.Contains(t, "deployment contains no instances", string(response))
+	})
+
+	t.Run("Deployment", func(t *testing.T) {
```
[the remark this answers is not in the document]
tonsV2: It does more than just that. This test is full walk-through and will basically try to do everything from creating, adding instance, deploying and finally destroying both instances and the deployment itself. I'm all in favor of giving it a more descriptive name but I'm not sure what to call it :-)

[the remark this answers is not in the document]
radnov: Some suggestions - `DeploymentAtoZ`, `DeploymentWalkthrough`, `DeploymentWorkflow`, `DeploymentLifecycle`, etc. I don't feel like any of them is the "correct" one, but they still give a bit more context. 

Maybe `Deployment` is just enough after all 🤷

radnov (approved):

radnov (approved):

PR #512, opened 2023-10-11
feat: initial instance status

tonsV2:


pkg/instance/service.go
```diff
@@ -661,6 +664,79 @@ func (s service) FindInstances(user *model.User, presets bool) ([]GroupsWithInst
...
+
+	pod, err := ks.getPod(instance, "")
+	if err != nil {
+		if strings.HasPrefix(err.Error(), "no pod found using the selector: ") {
+			return NotDeployed, nil
+		}
+		return "", err
+	}
+
+	if pod.Status.Phase == v1.PodPending {
```
radnov, line 692: Maybe use a `switch` statement instead, as we're checking the pod status for 3 different cases (and might have more?).

pkg/instance/service.go
```diff
@@ -661,6 +664,79 @@ func (s service) FindInstances(user *model.User, presets bool) ([]GroupsWithInst
...
+	if err != nil {
+		return "", err
+	}
+
+	pod, err := ks.getPod(instance, "")
+	if err != nil {
+		if strings.HasPrefix(err.Error(), "no pod found using the selector: ") {
+			return NotDeployed, nil
+		}
+		return "", err
```
radnov, line 689: Wrap this error? `fmt.Errorf("failed to get pod: %v", err)`

pkg/instance/service.go
```diff
@@ -661,6 +664,79 @@ func (s service) FindInstances(user *model.User, presets bool) ([]GroupsWithInst
...
+	Booting            InstanceStatus = "Booting"
+	BootingWithRestart InstanceStatus = "Booting (%d)"
+	Running            InstanceStatus = "Running"
+	Error              InstanceStatus = "Error"
+)
+
+func (s service) GetStatus(instance *model.Instance) (InstanceStatus, error) {
+	ks, err := NewKubernetesService(instance.Group.ClusterConfiguration)
+	if err != nil {
+		return "", err
```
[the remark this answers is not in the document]
tonsV2: No, the goal of such would be to encapsulate any Kubernetes client errors and that's already done from within `NewKubernetesService`.

pkg/instance/service.go
tonsV2: I updated the code but I'm not sure I find it more readable. Please take another look and tell me which version you like the most :smile:

pkg/instance/service.go
tonsV2: The error is already wrapped in the `getPod` method. The only reason for having the if statement here is because we don't want to return an error if the pod isn't found. It could simply be that it's not deployed.

I was wondering if it would be better to return errdef.NotFound rather than checking the error message though.

radnov: > I was wondering if it would be better to return errdef.NotFound rather than checking the error message though.

I think that makes sense, yeah.

pkg/instance/service.go
radnov: I suppose readability is the same with both options (at least for me). Choose whichever you prefer 👍

pkg/instance/service.go
```diff
@@ -661,6 +663,82 @@ func (s service) FindInstances(user *model.User, presets bool) ([]GroupsWithInst
...
+			if containerError {
+				status := fmt.Sprintf(string(BootingWithRestart), pod.Status.ContainerStatuses[0].RestartCount)
+				return InstanceStatus(status), nil
+			}
+
+			return Booting, nil
+		}
+		return Running, nil
+	}
+	return "", fmt.Errorf("unable to get instance status")
```
radnov, line 739: `failed to ...` for consistency

pkg/instance/kubernetes.go
```diff
@@ -158,7 +160,7 @@ func (ks kubernetesService) getPod(instance *model.Instance, typeSelector string
 	}
 
 	if len(pods.Items) == 0 {
-		return v1.Pod{}, fmt.Errorf("no pod found using the selector: %q", selector)
+		return v1.Pod{}, errdef.NewNotFound("no pod found using the selector: %q", selector)
```
radnov, line 163: `failed to ...` for consistency

radnov (approved):
LGTM 👍  

Added two minor comments about error message consistency.

PR #513, opened 2023-10-11
feat: prioritize parameters

tonsV2:
The goal of this PR is to ensure the stack paramters are shown in a consistent order.


radnov (approved):
Just a few comments/suggestions:

* I would use a more granular numbers for the priority (in the `10`  even `100` range), so that we don't have to change the order for the whole set of parameters if we want to change the order for one or add a new one.
* `IMAGE_PULL_POLICY` shouldn't have such a high priority IMO, it will just be confusing to most people. Currently the order is also nice and kind of categorized, but I think `IMAGE_TAG`, `IMAGE_REPOSITORY` and `DATABASE_ID` should always be the first 3.

tonsV2:
> Just a few comments/suggestions:
> 
> * I would use a more granular numbers for the priority (in the `10`  even `100` range), so that we don't have to change the order for the whole set of parameters if we want to change the order for one or add a new one.

I did that in the beginning but then just made it sequential. How about we change it if needed?

> * `IMAGE_PULL_POLICY` shouldn't have such a high priority IMO, it will just be confusing to most people. Currently the order is also nice and kind of categorized, but I think `IMAGE_TAG`, `IMAGE_REPOSITORY` and `DATABASE_ID` should always be the first 3.

I agree with you point of view but I felt `IMAGE_PULL_POLICY` belonged `IMAGE_TAG` and `IMAGE_REPOSITORY` since they're kinda related. Once people start using this for development they will change the policy unless they use a new tag for each build

PR #565, opened 2023-11-09
Bump versions

tonsV2:
The goal of this PR is to bump versions pieces of software we're using. As part of such we're starting to use the default "slices" package rather than the experimental one which take a different argument.


skaffold.yaml
```diff
@@ -29,7 +29,7 @@ deploy:
         createNamespace: true
         remoteChart: rabbitmq
         repo: https://charts.bitnami.com/bitnami
-        version: 10.3.9
+        version: 12.2.5
```
radnov, line 32: There's a newer version `12.4.1` now, should we use it?

skaffold.yaml
```diff
@@ -39,7 +39,7 @@ deploy:
         createNamespace: true
         remoteChart: redis
         repo: https://charts.bitnami.com/bitnami
-        version: 17.6.0
+        version: 18.2.1
```
radnov, line 42: There's a newer version `18.2.2` now, should we use it?

radnov (commented):
Should we also bump the [bitnami postgresql chart](https://artifacthub.io/packages/helm/bitnami/postgresql) version?

tonsV2:
> Should we also bump the [bitnami postgresql chart](https://artifacthub.io/packages/helm/bitnami/postgresql) version?

Postgresql requires a bit more work because we need to upgrade the existing data. So I deliberately left that out for now.

I've created [this](https://dhis2.atlassian.net/browse/DEVOPS-318) Jira task so we won't forget.

skaffold.yaml
tonsV2: Most certainly! I guess I forget to update my helm repo before searching, thanks :pray:

skaffold.yaml
tonsV2: Same as above

radnov (approved):

PR #584, opened 2023-11-16
docs: how to add a group

tonsV2:


README.md
```diff
@@ -1,3 +1,9 @@
 # TODO
 
 * Write readme
+
+# Add a group
+* Add group in IM (either through the UI or by using the user script scripts/users/createGroup.sh)
```
radnov, line 6: code formatting for the script path would be nice

README.md
```diff
@@ -1,3 +1,9 @@
...
+* Update [values file](https://github.com/dhis2-sre/im-manager/blob/8cb9a5959e334b835188fa07e801996ff2410b7c/helm/chart/values.yaml#L12) or for an individual environment such as [prod](https://github.com/dhis2-sre/im-manager/blob/8cb9a5959e334b835188fa07e801996ff2410b7c/helm/data/values/prod/values.yaml#L1).
```
radnov, line 7: Should use a URL that points to `master`, not a specific commit (applies to all URLs below, too). 

And should we even suggest making changes to the main values file of the chart?

radnov (changes requested):

README.md
tonsV2: > Should use a URL that points to `master`, not a specific commit (applies to all URLs below, too).

With a permanent link there's no doubt about what we're referring to. If things move around it should be easy to find in the lastest version since the user knows exactly what they're looking for.

> And should we even suggest making changes to the main values file of the chart?

Depends on the use case I suppose. I'm just making aware of it.

radnov (approved):

PR #619, opened 2023-12-04
refactor: delete external downloads before deleting database

tonsV2:
# When deleting a database the following behavior is now implemented
* If a database is locked and error message is returned
* All external downloads are deleted 

# Referential integrity
The Lock model now uses foreign keys so that the Instance, and user, it references has to exists

# Tests
* Lock
* Unlock
* ExternalDownloads
* Delete


pkg/database/database_integration_test.go
```diff
@@ -135,9 +166,64 @@ func TestDatabaseHandler(t *testing.T) {
...
+		// Attempt delete but expect a bad request response
+		client.Do(t, http.MethodDelete, "/databases/"+databaseID, nil, http.StatusBadRequest, inttest.WithAuthToken("sometoken"))
+		// Unlock database
+		client.Delete(t, "/databases/"+databaseID+"/lock")
+
+		// Create external download to ensure the database can still be deleted
+		client.PostJSON(t, "/databases/"+databaseID+"/external", strings.NewReader(`{
+				"expiration": 60
+			}`), &model.ExternalDownload{})
+
```
radnov, line 226: Should deleting a locked db or one with external downloads be split into subtests?

radnov (approved):

pkg/database/database_integration_test.go
tonsV2: Good point! I was considering it but it'll require quite a lot of boiler plate code so I just piggy backed the clean up code for now.

PR #625, opened 2023-12-05
ci: trigger im-web-client e2e tests workflow

radnov:
Related to https://github.com/dhis2-sre/im-web-client/pull/352

This PR updates the e2e workflow to trigger the Playwright tests in the `im-web-client` repository, instead of using the `stack_tests.sh` script. The workflow will be triggered only if the `deploy` label is added and after the build-test-deploy workflow has completed, as we want to have the changes deployed before we trigger the e2e test in the `im-web-client` repository.

The PR also adds a default user, like the admin user we already have, that is added to the whoami group and will be used for testing.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -71,4 +71,5 @@ GROUP_HOSTNAMES=whoami.im.c.127.0.0.1.nip.io
 ADMIN_USER_EMAIL=andreas@dhis2.org
 ADMIN_USER_PASSWORD=somepassword
 
-S3_BUCKET=test-db-manager-bucket
+DEFAULT_USER_EMAIL=im-default@dhis2.org
+DEFAULT_USER_PASSWORD=somepassword
--- a/.env.example
+++ b/.env.example
@@ -71,5 +71,5 @@ GROUP_HOSTNAMES=whoami.im.c.127.0.0.1.nip.io
 ADMIN_USER_EMAIL=andreas@dhis2.org
 ADMIN_USER_PASSWORD=somepassword
 
-DEFAULT_USER_EMAIL=im-default@dhis2.org
-DEFAULT_USER_PASSWORD=somepassword
+E2E_TEST_USER_EMAIL=im-e2e-test@dhis2.org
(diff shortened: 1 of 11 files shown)


.env.example
```diff
@@ -71,4 +71,5 @@ GROUP_HOSTNAMES=whoami.im.c.127.0.0.1.nip.io
 ADMIN_USER_EMAIL=andreas@dhis2.org
 ADMIN_USER_PASSWORD=somepassword
 
-S3_BUCKET=test-db-manager-bucket
+DEFAULT_USER_EMAIL=im-default@dhis2.org
```
tonsV2, line 74: I don't like having a generic user like this and default doesn't really reveal it's purpose. Let's call it what it is... im-e2e-test@dhis2.org or such.

cmd/serve/main.go
```diff
@@ -157,14 +157,20 @@ func run() error {
 
 	integrationHandler := integration.NewHandler(dockerHubClient, cfg.InstanceService.Host, cfg.DatabaseManagerService.Host)
 
-	err = user.CreateAdminUser(cfg.AdminUser.Email, cfg.AdminUser.Password, userService, groupService)
+	//err = user.CreateAdminUser(cfg.AdminUser.Email, cfg.AdminUser.Password, userService, groupService)
```
tonsV2, line 160: Delete this comment?

cmd/serve/main.go
```diff
@@ -157,14 +157,20 @@ func run() error {
...
+	err = user.CreateUser(cfg.AdminUser.Email, cfg.AdminUser.Password, userService, groupService, model.AdministratorGroupName, "admin")
 	if err != nil {
 		return err
 	}
 	err = createGroups(cfg, groupService)
 	if err != nil {
 		return err
 	}
+	//err = user.CreateDefaultUser(cfg.DefaultUser.Email, cfg.DefaultUser.Password, userService, groupService)
```
tonsV2, line 169: Same here?

tonsV2 (approved):

PR #630, opened 2023-12-11
Encryption

tonsV2:
The goal of this PR is to implement encryption of parameters for deployment instances.

To avoid further code duplication this PR will disable encryption for the regular (non-deployment) instances.

### DIFF
--- a/pkg/instance/handler.go
+++ b/pkg/instance/handler.go
@@ -71,8 +71,8 @@ func (h Handler) DeployDeployment(c *gin.Context) {
 		return
 	}
 
-	canRead := handler.CanWriteDeployment(user, deployment)
-	if !canRead {
+	canWrite := handler.CanWriteDeployment(user, deployment)
+	if !canWrite {
 		unauthorized := errdef.NewUnauthorized("write access denied")
 		_ = c.Error(unauthorized)
 		return
--- a/pkg/instance/handler.go
+++ b/pkg/instance/handler.go
@@ -71,9 +71,9 @@ func (h Handler) DeployDeployment(c *gin.Context) {
 		return
 	}
 
-	canRead := handler.CanReadDeployment(user, deployment)
+	canRead := handler.CanWriteDeployment(user, deployment)
 	if !canRead {
-		unauthorized := errdef.NewUnauthorized("read access denied")
+		unauthorized := errdef.NewUnauthorized("write access denied")
 		_ = c.Error(unauthorized)
 		return
 	}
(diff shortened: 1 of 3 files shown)


pkg/instance/handler.go
```diff
@@ -65,15 +65,15 @@ func (h Handler) DeployDeployment(c *gin.Context) {
...
 
-	deployment, err := h.instanceService.FindDeploymentById(id)
+	deployment, err := h.instanceService.FindDecryptedDeploymentById(id)
 	if err != nil {
 		_ = c.Error(err)
 		return
 	}
 
-	canRead := handler.CanReadDeployment(user, deployment)
+	canRead := handler.CanWriteDeployment(user, deployment)
```
radnov, line 74: Should the variable here be `canWrite`?

pkg/instance/repository.go
```diff
@@ -75,8 +75,61 @@ func (r repository) FindDeploymentById(id uint) (*model.Deployment, error) {
...
 
+func (r repository) FindDeploymentInstanceById(id uint) (*model.DeploymentInstance, error) {
+	var instance *model.DeploymentInstance
+	err := r.db.
+		Preload("GormParameters").
+		First(&instance, id).Error
+
+	if err != nil {
+		if errors.Is(err, gorm.ErrRecordNotFound) {
+			return nil, errdef.NewNotFound("instance not found by id: %d", id)
```
radnov, line 86: Should we also use the standard `"failed to ..."` error message prefix here?

pkg/instance/repository.go
```diff
@@ -132,16 +183,16 @@ func (r repository) Unlink(instance *model.Instance) error {
...
-		return err
-	}
-
-	err = r.db.Session(&gorm.Session{FullSaveAssociations: true}).Save(instance).Error
+	/*
+		err := encryptParameters(key, instance)
+		if err != nil {
+			return err
+		}
+	*/
```
radnov, line 194: Are the commented out lines here intentional?

radnov (commented):
Added a few comments 👍 

Note that I have not tested the changes.

pkg/instance/repository.go
tonsV2: The entire method is to be deleted

pkg/instance/repository.go
radnov: Yeah, there's a mix of `X not found ...` and `failed to ...` error messages in both new and old code. Only the latter is mentioned in [the style docs here](https://github.com/dhis2-sre/im-doc/tree/master/dev#errors), so we should either stick to it (and even go back and apply it across all error messages) or update the doc. 👀

pkg/instance/repository.go
radnov: If it should be merged like this I'd add a `// TODO ...` above those lines.

radnov (approved):

PR #646, opened 2023-12-19
feat: implement TTL and public flag for deployments

tonsV2:


stacks/dhis2-core/helmfile.yaml
```diff
@@ -34,6 +34,8 @@ releases:
       - commonLabels:
           im: "true"
           im-default: "true"
+          im-deployment-id: "{{ requiredEnv "DEPLOYMENT_ID" }}"
+          im-instance-id: "{{ requiredEnv "INSTANCE_ID" }}"
           im-id: "{{ requiredEnv "INSTANCE_ID" }}"
```
radnov, line 39: Should we keep both `im-id` and `im-instance-id`, given that they will be the same?

radnov (approved):

PR #694, opened 2024-02-21
chore: update scripts to use deployments

tonsV2:


scripts/instances/deploy-dhis2.sh
```diff
@@ -4,14 +4,21 @@ set -euo pipefail
...
 NAME=$2
 shift
 shift
 DESCRIPTION=${*:-""}
 
+INSTANCE_TTL=${INSTANCE_TTL:-0}
+PUBLIC=${PUBLIC:-false}
+
+# TODO: Update with slug matching prod
+DATABASE_ID=${DATABASE_ID:-whoami-2-40-2-sql-gz}
```
radnov, line 17: That should be `test-dbs-sierra-leone-2-40-2-sql-gz`, right?

radnov (approved):

PR #719, opened 2024-03-19
feat: cookie based authentication

tonsV2:
The goal of this PR is to allow browser based application to use cookies for authentication.

The implementation still accepts access tokens through the http authentication header so non-browser based clients can operate as usually.

Issues:
* CI/CD needs to be updated and populate UI_URL before this PR is merged
* Why does seed.sh only work with cookies and not with the authorization http header?


internal/middleware/authentication.go
```diff
@@ -67,60 +70,38 @@ func (m AuthenticationMiddleware) TokenAuthentication(c *gin.Context) {
...
 }
 
-func extractUser(userData any) (*model.User, error) {
-	userMap, ok := userData.(map[string]any)
+func extractUser(token jwt.Token) (*model.User, error) {
+	userData, ok := token.Get("user")
 	if !ok {
-		return nil, errors.New("failed to parse user data")
+		return nil, errors.New("user not found in claims")
 	}
```
radnov, line 97: Should we keep the error text prefix be "failed to ..." here for consistency?

pkg/user/handler.go
```diff
@@ -183,14 +192,30 @@ func (h Handler) RefreshToken(c *gin.Context) {
...
-		_ = c.Error(err)
+	var refreshTokenString string
+
+	cookie, err := c.Cookie("refreshToken")
+	if err == nil {
+		refreshTokenString = cookie
+	}
+
+	// TODO: what if err isn't http.ErrNoCookie
+	if errors.Is(err, http.ErrNoCookie) {
```
radnov, line 204: Checking the error type after checking whether the error is nil doesn't make sense to me. Maybe the order needs to be changed or not even check if there's no error. 🤔

radnov (commented):
Looks okay to me, but left two comments, one of which at least needs clarification before approving.

radnov (approved):

internal/middleware/authentication.go
tonsV2: No :slightly_smiling_face:

internal/server/engine.go
```diff
@@ -1,18 +1,23 @@
...
-func GetEngine(basePath string) *gin.Engine {
+func GetEngine(basePath string, allowedOrigin string) *gin.Engine {
 	r := gin.Default()
 
 	corsConfig := cors.DefaultConfig()
-	corsConfig.AllowAllOrigins = true
+	// TODO: Allow multiple origins?
+	// Without specifying origin, secure cookies won't work
+	log.Println("Allowed origin:", allowedOrigin)
+	corsConfig.AllowOrigins = []string{allowedOrigin, "http://localhost:5173", "http://localhost:3000"}
```
tonsV2, line 20: There are some issues here which I'm not entirely sure how to deal with
1. We shouldn't hard code the two localhost url's
2. It's not going to work on http anyway. If we're using http the cookie should be marked as unsecure

radnov (approved):

internal/middleware/authentication.go
```diff
@@ -67,60 +68,38 @@ func (m AuthenticationMiddleware) TokenAuthentication(c *gin.Context) {
...
 
-	id := userMap["id"].(float64)
-	email := userMap["email"].(string)
-
-	user := &model.User{
-		ID:          uint(id),
-		Email:       email,
-		Groups:      extractGroups("groups", userMap),
-		AdminGroups: extractGroups("adminGroups", userMap),
+	bytes, err := json.Marshal(userData)
```
teleivo, line 97: Why do we serialize to Json and deserialize it right away?

internal/server/engine.go
```diff
@@ -1,18 +1,22 @@
...
 )
 
-func GetEngine(basePath string) *gin.Engine {
+func GetEngine(basePath string, allowedOrigins []string) *gin.Engine {
 	r := gin.Default()
 
 	corsConfig := cors.DefaultConfig()
-	corsConfig.AllowAllOrigins = true
+	// Without specifying origin, secure cookies won't work
+	log.Println("Allowed origin:", allowedOrigins)
```
teleivo, line 18: Why do we log this? We control the config so we know it right?

pkg/token/helper/tokenHelper.go
```diff
@@ -101,7 +100,7 @@ func GenerateRefreshToken(user *model.User, secretKey string, expirationInSecond
 	currentTime := time.Now()
 	tokenExpiration := currentTime.Add(time.Duration(expirationInSeconds) * time.Second)
 
-	tokenId, err := uuid.NewV4()
+	tokenId, err := uuid.NewUUID()
```
teleivo, line 103: You can use `uuid.NewString()` https://pkg.go.dev/github.com/google/uuid#NewString as you only want a string.

skaffold.yaml
```diff
@@ -9,6 +9,9 @@
     serviceAccount:
       name: im-manager-prod
       namespace: instance-manager-prod
+#    setValueTemplates:
+#      corsAllowOrigins:
+#        - "https://{{ .UI_URL }}"
```
teleivo, line 14: Why is that commented out? I would only add commented out code with an explanation. I have no clue why this is here and if I can delete it or when I should uncomment it 🤷🏻

pkg/user/handler_test.go
```diff
@@ -0,0 +1,238 @@
...
+	"github.com/dhis2-sre/im-manager/pkg/model"
+	"github.com/dhis2-sre/im-manager/pkg/token"
+	"github.com/gin-gonic/gin"
+	"github.com/google/uuid"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/require"
+)
+
+func TestHandler_RefreshToken_Cookie(t *testing.T) {
```
teleivo, line 20: Why did you not add a test to user_integration_test.go? Mock based tests don't give me much confidence it actually works which for something involving users and auth is key 😅

helm/chart/values.yaml
```diff
@@ -2,6 +2,10 @@
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
 
+corsAllowOrigins: []
+#  - http://localhost:3000
```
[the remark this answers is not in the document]
tonsV2: I guess we should supply a default value

cmd/serve/main.go
```diff
@@ -83,7 +83,7 @@ func run() error {
 		return err
 	}
 
-	userHandler := user.NewHandler(userService, tokenService)
+	userHandler := user.NewHandler(cfg, userService, tokenService)
```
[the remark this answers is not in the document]
teleivo: can you pass hostname and authconfig at least instead of the entire config?

internal/middleware/authentication.go
teleivo: I am not exactly sure what `userData` is. Is it a JSON string? Or does is it actually a `*model.User`? In the latter case we can type assert. If its not maybe https://github.com/lestrrat-go/jwx/pull/391 can help 🤔

tonsV2: Yes! Good find!

PR #724, opened 2024-03-27
feat: remember me

tonsV2:


pkg/token/service.go
```diff
@@ -56,27 +54,30 @@ type RefreshTokenData struct {
...
-func (t tokenService) GetTokens(user *model.User, previousRefreshTokenId string) (*Tokens, error) {
+func (t tokenService) GetTokens(user *model.User, previousRefreshTokenId string, rememberMe bool) (*Tokens, error) {
 	if previousRefreshTokenId != "" {
 		if err := t.repository.DeleteRefreshToken(user.ID, previousRefreshTokenId); err != nil {
 			return nil, errdef.NewUnauthorized("could not delete previous refreshToken for user.Id: %d, tokenId: %s", user.ID, previousRefreshTokenId)
 		}
 	}
 
-	accessToken, err := helper.GenerateAccessToken(user, t.privateKey, t.accessTokenExpirationSeconds)
+	accessToken, err := helper.GenerateAccessToken(user, t.privateKey, t.authentication.AccessTokenExpirationSeconds)
```
teleivo, line 70: I think our code would benefit from using `time.Duration` for such values. We could stick to the `_SECONDS` env vars but then translate into the `time.Duration` type when these configs leave the config package. `time.Duration` is easier to work with when computing things like a deadline, ... and since its typed we can shorten our var names 😋

pkg/user/handler.go
```diff
@@ -297,23 +306,34 @@ func (h Handler) RefreshToken(c *gin.Context) {
...
 	}
 
-	tokens, err := h.tokenService.GetTokens(user, refreshToken.ID.String())
+	rememberMe := false
+	rememberMeCookie, _ := c.Cookie("rememberMe")
+	if rememberMeCookie == "true" {
+		rememberMe = true
+	}
+
+	tokens, err := h.tokenService.GetTokens(user, refreshToken.ID.String(), rememberMe)
```
teleivo, line 315: I think we should pass in a `UUID` instead of a `string`. This way we remove a whole bunch of invalid strings from the possible inputs.

pkg/user/handler.go
```diff
@@ -363,6 +383,9 @@ func (h Handler) SignOut(c *gin.Context) {
 	//	200:
 	//	401: Error
 	//	415: Error
+
+	// TODO: cookies should be cleared no matter what...
```
teleivo, line 387: Can we do this correctly now?

pkg/user/handler.go
```diff
@@ -124,7 +124,7 @@ func (h Handler) ValidateEmail(c *gin.Context) {
 	//   404: Error
 	var request validateEmailRequest
 	if err := handler.DataBinder(c, &request); err != nil {
-		_ = c.Error(handleSignUpErrors(err))
+		_ = c.Error(errdef.NewBadRequest("validate email error: %v", err))
```
radnov, line 127: The wording seems a bit off. Maybe `failed to validate email ...` ?

pkg/token/service.go
tonsV2: As much as I agree I'd say it's beyond the scope of this PR to change our current pattern

pkg/user/handler.go
tonsV2: I agree and also tried to change it but two things...

1. It's not something I introduced as part of this PR
2. The change will introduce some not so nice code when refreshing the token

Due to the above I'd appreciate keeping it out of this PR

pkg/user/handler.go
tonsV2: The reason for this comment is that the endpoint is authenticated which is because if we need the user id to delete its associated tokens from Redis. So if the token is already expired we're not clearing the cookies. I'm not sure what the best way to go about this is.

pkg/user/handler.go
```diff
@@ -23,22 +21,26 @@ import (
...
-	hostname       string
-	authentication config.Authentication
-	publicKey      *rsa.PublicKey
-	userService    userService
-	tokenService   tokenService
+	hostname                                string
+	accessTokenExpirationSeconds            int
+	refreshTokenExpirationSeconds           int
+	refreshTokenRememberMeExpirationSeconds int
+	publicKey                               *rsa.PublicKey
```
teleivo, line 41: avoid using a pointer as we should not be able to mutate the public key here

tonsV2: The reason for using a point is that rsa.GenerateKey, rsa.GenerateKey etc. all return pointers. You mentioned this before and if I recall we even have a Jira issue. It does make sense to prevent mutation so I created [this](https://github.com/dhis2-sre/im-manager/pull/724/commits/7ad16a65109033e9a40425d2acd6e148d3c088c1) commit. Please let me know what you think :slightly_smiling_face:

tonsV2: As mentioned on Slack the commit has been removed and we'll deal with this later

pkg/user/user_integration_test.go
```diff
@@ -84,20 +76,102 @@ func TestUserHandler(t *testing.T) {
...
+
+			require.Equal(t, http.StatusUnauthorized, response.StatusCode)
+			assert.EqualValues(t, expectedCookies, response.Cookies())
+		})
+
+		t.Run("ValidToken", func(t *testing.T) {
+			createUser(t, client, userService, "user4@dhis2.org", "oneoneoneoneoneoneone111")
+
+			var tokens *token.Tokens
+			client.PostJSON(t, "/tokens", strings.NewReader(`{}`), &tokens, inttest.WithBasicAuth("user4@dhis2.org", "oneoneoneoneoneoneone111"))
```
teleivo, line 147: is passing `strings.NewReader(`{}`)` necessary?
we did have working requests using no body so passing `nil` before

pkg/user/user_integration_test.go
```diff
@@ -303,6 +591,30 @@ func TestUserHandler(t *testing.T) {
...
+	ValidateEmail(emailToken uuid.UUID) error
+}
+
+func createUser(t *testing.T, client *inttest.HTTPClient, userService userService, email string, password string) *model.User {
+	requestBody := strings.NewReader(fmt.Sprintf(`{
+			"email":    "%s",
+			"password": "%s"
+		}`, email, password))
+
+	user := model.User{}
```
teleivo, line 605: use `var user model.User` https://github.com/dhis2-sre/im-doc/blob/master/dev/README.md#variable-declaration

pkg/user/user_integration_test.go
```diff
@@ -303,6 +591,30 @@ func TestUserHandler(t *testing.T) {
 	})
 }
 
+type userService interface {
+	FindById(context context.Context, id uint) (*model.User, error)
+	ValidateEmail(emailToken uuid.UUID) error
+}
+
+func createUser(t *testing.T, client *inttest.HTTPClient, userService userService, email string, password string) *model.User {
```
teleivo, line 599: call https://pkg.go.dev/testing#T.Helper in helpers so we get better error messages

pkg/user/user_integration_test.go
```diff
@@ -84,20 +76,102 @@ func TestUserHandler(t *testing.T) {
...
+			assert.EqualValues(t, expectedCookies, response.Cookies())
+		})
+
+		t.Run("ExpiredToken", func(t *testing.T) {
+			createUser(t, client, userService, "user5@dhis2.org", "oneoneoneoneoneoneone111")
+
+			var tokens *token.Tokens
+			client.PostJSON(t, "/tokens", strings.NewReader(`{}`), &tokens, inttest.WithBasicAuth("user5@dhis2.org", "oneoneoneoneoneoneone111"))
+
+			time.Sleep(time.Duration(tokens.ExpiresIn) * time.Second)
```
teleivo, line 164: Are we now waiting 10 seconds? Maybe we can shrink that config for the test. I think https://pkg.go.dev/time#After would be better so the runtime gets a chance to maybe run another test while we wait here.

`<-time.After(time.Duration(tokens.ExpiresIn) * time.Second)`

pkg/user/user_integration_test.go
```diff
@@ -156,7 +230,9 @@ func TestUserHandler(t *testing.T) {
 			{
 				t.Log("SignIn")
 
-				client.PostJSON(t, "/tokens", nil, &user1Token, inttest.WithBasicAuth("user1@dhis2.org", "oneoneoneoneoneoneone111"))
+				createUser(t, client, userService, "user6@dhis2.org", "oneoneoneoneoneoneone111")
```
teleivo, line 233: Why do we need a different user here and in `GetMe`?

pkg/user/user_integration_test.go
```diff
@@ -167,14 +243,196 @@ func TestUserHandler(t *testing.T) {
...
+
+				createUser(t, client, userService, "user7@dhis2.org", "oneoneoneoneoneoneone111")
+				requestBody := strings.NewReader(`{}`)
+				request := client.NewRequest(t, http.MethodPost, "/tokens", requestBody, inttest.WithBasicAuth("user7@dhis2.org", "oneoneoneoneoneoneone111"), inttest.WithHeader("Content-Type", "application/json"))
+
+				response, err := client.Client.Do(request)
+				require.NoError(t, err)
+
+				assert.Equal(t, http.StatusCreated, response.StatusCode)
+				actualCookies := response.Cookies()
```
teleivo, line 266: Can we use something like `assert.EqualValues(t, expectedCookies, response.Cookies())`? So instead of asserting on individual fields assert on a struct. This makes the test easier to read. It also prevents the whac-a-mole of I fixed one failing assertion, rerun only to find another field was also not equal, ...

pkg/user/user_integration_test.go
teleivo, line 599: You are returning a user but not using it. You could generate the `email` and 
`password` in here, save 2 params and return and use the user to create the token in the requests.

Maybe use a uuid as in `uuid@dhis2.org` for email and password.

It might even be nice to create a struct which has an https://pkg.go.dev/sync/atomic#Int32 field. Then this method increments it on every call. This would be nice as I see we have one test that asserts how many users are returned which is a magic number.

pkg/user/user_integration_test.go
```diff
@@ -41,40 +45,28 @@ func TestUserHandler(t *testing.T) {
...
+		userHandler := user.NewHandler("hostname", 10, 20, 30, &key.PublicKey, userService, tokenService)
 		user.Routes(engine, authentication, authorization, userHandler)
 	})
 
 	var user1ID string
 	{
 		t.Log("SignUpUsers")
 
-		var user1 model.User
-		client.PostJSON(t, "/users", strings.NewReader(`{
```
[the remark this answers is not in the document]
tonsV2: I noticed that we actually didn't test the validate email endpoint either so I added that as well

pkg/user/user_integration_test.go
tonsV2: The difference is that now the endpoint accepts an optional request body.

In the handler I can do the following

	var request signInRequest
	if err := handler.DataBinder(c, &request); err != nil && err != io.EOF {
		_ = c.Error(err)
		return
	}

The relevant part is the last condition which means we accept io.EOF errors. Can you think of a better solution?

pkg/user/user_integration_test.go
tonsV2: It's difficult because `Value` and `Raw` contains the token and I don't know what it is

pkg/user/user_integration_test.go
tonsV2: Please see the updated version

pkg/user/user_integration_test.go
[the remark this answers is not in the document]
teleivo: No tests added in here back when the feature was added https://github.com/dhis2-sre/im-manager/pull/393

pkg/user/user_integration_test.go
teleivo: So you are saying you want to assert only on some but not other fields? testify does not seem to be able to do that 😢 

https://github.com/google/go-cmp can which is my preferred tool when asserting.

Then at least create a helper like

`assertCookie(t *testing.T, ....)`

This makes the test more readable as we have tons of lines just for asserting cookies 😅

pkg/user/user_integration_test.go
```diff
@@ -290,19 +575,61 @@ func TestUserHandler(t *testing.T) {
...
+
+func createUser(t *testing.T, client *inttest.HTTPClient, userService userService) (uint, string, string) {
+	t.Helper()
+
+	userCounter.Add(1)
+	currentUserCount := userCounter.Load()
+
+	email := fmt.Sprintf("user%d@dhis2.org", currentUserCount)
+
+	passwordLength := 24
```
teleivo, line 608: isn't a uuid a valid password? if it is its less code 😅

pkg/user/user_integration_test.go
```diff
@@ -290,19 +575,61 @@ func TestUserHandler(t *testing.T) {
 			var users []model.User
 			client.GetJSON(t, "/users", &users, inttest.WithAuthToken(adminToken.AccessToken))
 
-			assert.Lenf(t, users, 4, "GET /users should return 4 users one of which is an admin")
+			assert.Lenf(t, users, 13, "GET /users should return 13 users one of which is an admin")
```
teleivo, line 578: We should get this number from our userCounter

pkg/user/user_integration_test.go
```diff
@@ -41,63 +44,145 @@ func TestUserHandler(t *testing.T) {
...
-		client.PostJSON(t, "/users", strings.NewReader(`{
-			"email":    "user2@dhis2.org",
-			"password": "oneoneoneoneoneoneone111"
-		}`), &user2)
+		t.Log("SignIn")
 
-		require.Equal(t, "user2@dhis2.org", user2.Email)
-		require.Empty(t, user2.Password)
+		var tokens *token.Tokens
+		client.PostJSON(t, "/tokens", strings.NewReader(`{}`), &tokens, inttest.WithBasicAuth("user@dhis2.org", "oneoneoneoneoneoneone111"))
```
teleivo, line 87: Is the `strings.NewReader(`{}`)` needed? `nil` worked just fine before

pkg/user/user_integration_test.go
```diff
@@ -152,99 +237,304 @@ func TestUserHandler(t *testing.T) {
...
 			}
+
+			{
+				t.Log("RefreshTokensRequestBody")
+
+				_, email, password := createUser(t, client, userService)
+				var tokens *token.Tokens
+				client.PostJSON(t, "/tokens", strings.NewReader(`{}`), &tokens, inttest.WithBasicAuth(email, password))
+				require.NotEmpty(t, tokens.AccessToken, "should return an access token")
+				requestBody := strings.NewReader(fmt.Sprintf(`{"refreshToken": "%s"}`, tokens.RefreshToken))
```
teleivo, line 419: since we are doing something like that often a helper like

```
func jsonBody(format string, args ...any) io.Reader {
  return strings.NewReader(fmt.Sprintf(format, args...))
}
```

(or name it just `body` as its agnostic to what you put in)

makes sense to have it read like

```
request := client.NewRequest(t, http.MethodPost, "/refresh", jsonBody(`{"refreshToken": "%s"}`, tokens.RefreshToken), inttest.WithHeader("Content-Type", "application/json"))
```

pkg/user/router.go
```diff
@@ -7,6 +7,7 @@ import (
 
 func Routes(r *gin.Engine, authenticationMiddleware middleware.AuthenticationMiddleware, authorizationMiddleware middleware.AuthorizationMiddleware, handler Handler) {
 	r.POST("/users", handler.SignUp)
+	r.DELETE("/users", handler.SignOut)
```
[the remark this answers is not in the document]
tonsV2: Let's say a user has checked the "Remember me" box when logging in but both the access token and the refresh token is expired. So when sending the logout request we'll simply reset the cookies rather than forcing the user to login again to do so.

pkg/user/handler.go
```diff
@@ -364,9 +394,14 @@ func (h Handler) SignOut(c *gin.Context) {
...
 	//	415: Error
-	user, err := handler.GetUserFromContext(c)
+
+	// No matter what happens, if the user sends a sign-out request, delete all cookies
+	unsetCookie(c)
+
+	user, err := h.parseRequest(c.Request)
 	if err != nil {
-		_ = c.Error(err)
+		log.Println("token not valid:", err)
```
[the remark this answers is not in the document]
tonsV2: So what are you suggestions

		h.logger.Info("token not valid", "error", err)
		_ = c.Error(errdef.NewUnauthorized("token not valid"))

Like so?

pkg/user/user_integration_test.go
tonsV2: I've committed the above. Let me know if you think it's a bad idea :slightly_smiling_face:

pkg/user/handler.go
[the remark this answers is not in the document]
teleivo: `h.logger.Error("token not valid", "error", err)` since its an error 😄

pkg/user/user_integration_test.go
tonsV2: Please see the conversation we're having further up

pkg/user/handler.go
```diff
@@ -247,7 +248,7 @@ func (h Handler) SignIn(c *gin.Context) {
 	}
 
 	var request signInRequest
-	if err := handler.DataBinder(c, &request); err != nil {
+	if err := handler.DataBinder(c, &request); err != nil && err != io.EOF {
```
teleivo, line 251: why is that needed?

tonsV2: Because that's the error which is returned if the request body is nil

pkg/user/user_integration_test.go
teleivo: > The difference is that now the endpoint accepts an optional request body.
if the body is optional than passing `nil` in the test should work

I think the solution you used in the handler is alright for now. We might need to rethink the `DataBinder` at some point though 😄

pkg/user/user_integration_test.go
```diff
@@ -573,7 +531,7 @@ func TestUserHandler(t *testing.T) {
 			var users []model.User
 			client.GetJSON(t, "/users", &users, inttest.WithAuthToken(adminToken.AccessToken))
 
-			expectedNumberOfUsers := int(userCounter.Load())
+			expectedNumberOfUsers := int(userCounter.Load()) + 3 // counter + 1 admin + 1 manually created + 1 manually created which isn't validated
```
teleivo, line 534: you could increment the counter at the sites where these users are create

teleivo (approved):

pkg/user/user_integration_test.go
tonsV2: True but not for the admin, I've updated the code

PR #730, opened 2024-04-11
feat: add reset and request reset password endpoints

radnov:


pkg/user/user_integration_test.go
```diff
@@ -51,6 +51,8 @@ func TestUserHandler(t *testing.T) {
 		user.Routes(engine, authentication, authorization, userHandler)
 	})
 
+	// TODO add tests for resetting password
```
tonsV2, line 54: Is this comment still relevant?

pkg/user/user_integration_test.go
```diff
@@ -206,6 +208,42 @@ func TestUserHandler(t *testing.T) {
...
+			t.Parallel()
+
+			{
+				t.Log("RequestPasswordReset")
+
+				requestResetRequestBody := strings.NewReader(`{
+					"email":    "user1@dhis2.org"
+				}`)
+
+				client.Do(t, http.MethodPost, "/users/request-reset", requestResetRequestBody, http.StatusOK, inttest.WithHeader("Content-Type", "application/json"))
```
tonsV2, line 222: Don't we have a client.PostJSON (or such) method?

tonsV2 (approved):

pkg/user/user_integration_test.go
radnov: We do, but [it expects a 201 response](https://github.com/dhis2-sre/im-manager/blob/03422fb6e388b7bf8dca7fdb7d10a6348420b399/pkg/inttest/http.go#L104), while we get a 200 here.

This begs the question - should we use a 201 response for both of the request password reset and reset password endpoints? 🤔

tonsV2: I would be ok with returning 201. But I'll leave it up to you how we proceed here :slightly_smiling_face:

radnov: I suppose it makes sense to return 201, as we are creating a password reset token and then updating the password. 🤔

pkg/model/user.go
```diff
@@ -9,15 +10,17 @@ import (
...
-	AdminGroups []Group   `json:"adminGroups" gorm:"many2many:user_groups_admin;constraint:OnUpdate:CASCADE,OnDelete:CASCADE"`
+	ID               uint           `json:"id" gorm:"primarykey"`
+	CreatedAt        time.Time      `json:"createdAt"`
+	UpdatedAt        time.Time      `json:"updatedAt"`
+	Email            string         `json:"email" gorm:"index;unique"`
+	EmailToken       uuid.UUID      `json:"-" gorm:"unique;type:uuid"`
+	Validated        bool           `json:"validated"`
+	Password         string         `json:"-"`
+	PasswordToken    sql.NullString `json:"-"`
+	PasswordTokenTTL uint           `json:"-"`
```
tonsV2, line 21: Why do we have this as a property on the user model?

pkg/user/handler.go
```diff
@@ -135,6 +137,69 @@ func (h Handler) ValidateEmail(c *gin.Context) {
...
+	err := h.userService.RequestPasswordReset(request.Email)
+	if err != nil {
+		_ = c.Error(err)
+		return
+	}
+
+	c.Status(http.StatusCreated)
+}
+
+type resetPasswordRequest struct {
```
tonsV2, line 171: Are you using this struct client side? When you construct the request, do you ensure this type is what's sent as the request body?

pkg/user/repository.go
```diff
@@ -115,3 +126,17 @@ func (r repository) update(user *model.User) (*model.User, error) {
...
 	return user, nil
 }
+
+func (r repository) resetPassword(user *model.User) (*model.User, error) {
+	updatedUser := model.User{
+		Password:      user.Password,
+		PasswordToken: sql.NullString{String: "", Valid: false},
+	}
+
+	err := r.db.Model(&user).Select("Password", "PasswordToken").Updates(updatedUser).Error
```
tonsV2, line 136: Most other places in the file we add new lines between the method calls for the sake of readability. I think that should be done here as well

pkg/user/service.go
```diff
@@ -78,7 +83,7 @@ func (s service) sendValidationEmail(user *model.User) error {
 	m.SetHeader("To", user.Email)
 	m.SetHeader("Subject", "Welcome to IM")
 	link := fmt.Sprintf("%s/validate/%s", s.config.UIHostname, user.EmailToken)
-	body := fmt.Sprintf("Hello, please click the below link to verify your email.<br/>%s", link)
+	body := fmt.Sprintf("Hello, please click the link below to verify your email.<br/>%s", link)
```
tonsV2, line 86: I think both sentences are valid English but I don't mind either way :wink:

pkg/user/service.go
```diff
@@ -205,3 +210,68 @@ func (s service) Update(id uint, email, password string) (*model.User, error) {
...
+	body := fmt.Sprintf("Hello, please click the link below to reset your password.<br/>%s", link)
+	m.SetBody("text/html", body)
+	return s.dailer.DialAndSend(m)
+}
+
+func (s service) RequestPasswordReset(email string) error {
+	user, err := s.repository.findByEmail(email)
+
+	if err != nil {
+		return nil
```
tonsV2, line 237: Do we really want to silently suppress all errors here?

tonsV2: I think we want to do something like [this](https://github.com/dhis2-sre/im-manager/pull/730/files#diff-5a3d7d2e1ba1813b40192dfd630c1919e3fcd8bc1436560e793ce7af2bd98068R216).

pkg/user/service.go
```diff
@@ -205,3 +210,68 @@ func (s service) Update(id uint, email, password string) (*model.User, error) {
...
+	}
+
+	bytes := make([]byte, 64)
+	if _, err := rand.Read(bytes); err != nil {
+		return err
+	}
+	token := base64.URLEncoding.EncodeToString(bytes)
+
+	user.PasswordToken = sql.NullString{String: token, Valid: true}
+	user.PasswordTokenTTL = uint(time.Now().Unix()) + s.config.PasswordTokenTTL
```
tonsV2, line 247: Do we need to store this on the user model?

pkg/user/user_integration_test.go
```diff
@@ -206,6 +207,59 @@ func TestUserHandler(t *testing.T) {
...
+			{
+				t.Log("RequestPasswordReset")
+
+				requestResetRequestBody := strings.NewReader(`{
+					"email":    "user1@dhis2.org"
+				}`)
+
+				client.Do(t, http.MethodPost, "/users/request-reset", requestResetRequestBody, http.StatusCreated, inttest.WithHeader("Content-Type", "application/json"))
+
+				user1IDInt, _ := strconv.ParseUint(user1ID, 10, 0)
```
tonsV2, line 223: require.NoError(...) ?

tonsV2: Everywhere `user1IDInt` is used, it's converted to a uint. Would it make sense just doing that here?

pkg/user/repository.go
```diff
@@ -115,3 +126,17 @@ func (r repository) update(user *model.User) (*model.User, error) {
 
 	return user, nil
 }
+
+func (r repository) resetPassword(user *model.User) (*model.User, error) {
```
tonsV2, line 130: We return the user, but do we actually use it for anything anywhere?

tonsV2 (changes requested):

pkg/user/user_integration_test.go
tonsV2: I actually also thought a bit more about it, and yes, 201 is the way to go :+1:

pkg/model/user.go
tonsV2: I understand now, it's the time of the request +  expiry in seconds

tonsV2 (approved):
We actually shouldn't pass the entire config object but rather just the two properties we're using

tonsV2 (approved):

PR #766, opened 2024-05-20
feat: user scripts with cookie auth

tonsV2:
This PR offers two improvements over the previous implementation of authentication
* The access token is retrieved from the cookie and not the response body. Soon enough we'll remove the response body token for the sake of security so this is required going forward
* Support both administrators and regular users by using a different cache file for each user type

This can be tested by executing the following commands

    ./me.sh
    ./findAllGroups.sh
    ./me.sh

The first command will show only one group while the second will show two. And the final command will also only show one.

If [this](https://github.com/dhis2-sre/im-manager/pull/766/files#diff-78239d76684f8871a629a835c7c5c48eed2bc09adf6815935eb87a231ca0d805R7) line is changed to not include the `$IM_USER_TYPE` all commands will include the same groups as the token from the first command grant access to.


radnov (approved):
Nice! 🚀

PR #777, opened 2024-06-04
chore: get public from private key

teleivo:


pkg/config/config.go
```diff
@@ -150,20 +146,18 @@ type redis struct {
...
 }
 
-type keys struct {
-	PrivateKey string
-	PublicKey  string
-}
-
-func (k keys) GetPrivateKey(logger *slog.Logger) (*rsa.PrivateKey, error) {
-	decode, _ := pem.Decode([]byte(k.PrivateKey))
+func GetPrivateKey(logger *slog.Logger) (*rsa.PrivateKey, error) {
```
tonsV2, line 155: Currently we're doing this at run time. Could we perhaps do this during "boot time"?

PR #813, opened 2024-06-21
feat: list public deployments

tonsV2:


swagger/swagger.yaml
```diff
@@ -1078,6 +1078,20 @@ paths:
             security:
                 - oauth2: []
             summary: Delete deployment instance
+    /deployments/public:
+        get:
+            description: Find all public deployments
+            operationId: listPublicDeployments
```
teleivo, line 1084:
```suggestion
            operationId: findPublicDeployments
```

Or call it List all public deployments

pkg/instance/instance_integration_test.go
```diff
@@ -207,6 +207,46 @@ func TestInstanceHandler(t *testing.T) {
 		time.Sleep(3 * time.Second)
 		k8sClient.AssertPodIsNotRunning(t, deploymentInstance.GroupName, deploymentInstance.Name)
 	})
+
+	t.Run("GetPublicDeployments", func(t *testing.T) {
+		t.Log("Create deployment")
```
teleivo, line 212: Can this run in parallel to the other tests? Or we merge this with the other tests we have so that we put the `t.Log("Create public deployment")` block in to `t.Run("Deployment", func(t *testing.T) {` and then the `t.Log("Get public deployments")` at the end of it. As we do with the user integration tests.

pkg/instance/instance_integration_test.go
```diff
@@ -207,6 +207,46 @@ func TestInstanceHandler(t *testing.T) {
...
+		var deployment model.Deployment
+		body := strings.NewReader(`{
+			"name": "private-deployment",
+			"group": "group-name",
+			"description": "some description"
+		}`)
+		client.PostJSON(t, "/deployments", body, &deployment, inttest.WithAuthToken("sometoken"))
```
teleivo, line 219: Please add newlines in between the setup, the thing we test and the assertions as we do in other tests.

```
setup

client.Post..

assert
```

pkg/instance/instance_integration_test.go
```diff
@@ -207,6 +207,46 @@ func TestInstanceHandler(t *testing.T) {
...
+		assert.Equal(t, "public-deployment", deployment.Name)
+		assert.Equal(t, "group-name", deployment.GroupName)
+		assert.Equal(t, "some description", deployment.Description)
+		assert.True(t, deployment.Public)
+
+		t.Log("Get public deployments")
+		var groupsWithDeployments []instance.GroupsWithDeployments
+		client.GetJSON(t, "/deployments/public", &groupsWithDeployments)
+		assert.Len(t, groupsWithDeployments, 1)
```
teleivo, line 241: Using a custom struct for the assertion and https://pkg.go.dev/github.com/stretchr/testify/assert#EqualValues might make it more readable

radnov (approved):
Nothing to add, lgtm

pkg/instance/repository.go
```diff
@@ -164,6 +164,16 @@ func (r repository) FindDeployments(groupNames []string) ([]*model.Deployment, e
 	return deployments, err
 }
 
+func (r repository) FindPublicDeployments() ([]*model.Deployment, error) {
```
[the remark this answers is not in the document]
tonsV2: Not without changing code that we're also using when returning deployments elsewhere. I'd like to have this PR merged since the frontenders depend on it. So I'd appreciate keeping it as simple as possible.

PR #844, opened 2024-07-13
feat: remove token from response body

tonsV2:
The goal of this PR is to remove the tokens from the response bodies of /tokens and /refresh how this is also means that we had to change some code related to how our tests works. Therefor I introduced a SignIn method on the HTTPClient struct which returns the two cookies.


pkg/inttest/http.go
```diff
@@ -171,6 +171,41 @@ func (hc *HTTPClient) PostJSON(t *testing.T, path string, requestBody io.Reader,
...
+	require.NoError(t, err)
+
+	refreshToken, err := findCookie(res, "refreshToken")
+	require.NoError(t, err)
+
+	return accessToken, refreshToken
+}
+
+func findCookie(res *http.Response, name string) (*http.Cookie, error) {
+	cookies := res.Cookies()
```
teleivo, line 200: Such vars do not add value (var is named like the method) and just increase the scope of the var

pkg/inttest/http.go
```diff
@@ -171,6 +171,41 @@ func (hc *HTTPClient) PostJSON(t *testing.T, path string, requestBody io.Reader,
 	require.NoError(t, err, errMsg+": failed to unmarshal response body")
 }
 
+func (hc *HTTPClient) SignIn(t *testing.T, email, password string) (*http.Cookie, *http.Cookie) {
+	t.Helper()
+
+	req := hc.NewRequest(t, http.MethodPost, "/tokens", nil, WithHeader("Content-Type", "application/json"), WithBasicAuth(email, password))
```
teleivo, line 177: Why do you pass `Content-Type: application/json` in this post? There is no body right 🤔

pkg/user/user_integration_test.go
```diff
@@ -322,11 +316,10 @@ func TestUserHandler(t *testing.T) {
 				t.Log("RefreshTokensUsingCookie")
 
 				_, email, password := createUser(t, client, userService)
-				var tokens *token.Tokens
-				client.PostJSON(t, "/tokens", nil, &tokens, inttest.WithBasicAuth(email, password))
-				require.NotEmpty(t, tokens.AccessToken, "should return an access token")
+				accessToken, refreshToken := client.SignIn(t, email, password)
+				require.NotEmpty(t, accessToken.Value, "should return an access token")
```
teleivo, line 320: Could we move this require inside of `client.SignIn`? I assume it should never be empty after signin right?

pkg/inttest/http.go
```diff
@@ -171,6 +171,41 @@ func (hc *HTTPClient) PostJSON(t *testing.T, path string, requestBody io.Reader,
...
+	res := hc.do(t, req)
+	require.Zero(t, res.ContentLength)
```
teleivo, line 179: some message saying what we expected is useful in those requires hidden in the client (same below with finding cookies)

pkg/user/user_integration_test.go
```diff
@@ -531,21 +521,21 @@ func TestUserHandler(t *testing.T) {
...
+			require.NotEmpty(t, adminAccessToken.Value, "should return an access token")
 		}
 
 		{
 			t.Log("GetAllUsers")
 			userCount.Wait()
 
 			var users []model.User
-			client.GetJSON(t, "/users", &users, inttest.WithAuthToken(adminToken.AccessToken))
+			client.GetJSON(t, "/users", &users, inttest.WithAuthToken(adminAccessToken.Value))
```
teleivo, line 538: Do we now always pass in a cookie? If so you could change the `WithAuthToken` or make it `WithCookie`

pkg/inttest/http.go
tonsV2: We only accept those two types of content.

https://github.com/dhis2-sre/im-manager/blob/master/internal/handler/dataBinder.go#L10

pkg/user/user_integration_test.go
tonsV2: We don't always pass in the a cookie. We still accept a bearer token using the authorization header

pkg/inttest/http.go
teleivo: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type
> In requests, (such as [POST](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/POST) or [PUT](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/PUT)), the client tells the server what type of data is actually sent. 

You do not send any data, so this header should not be needed, right?

pkg/inttest/http.go
```diff
@@ -176,7 +176,7 @@ func (hc *HTTPClient) SignIn(t *testing.T, email, password string) (*http.Cookie
 
 	req := hc.NewRequest(t, http.MethodPost, "/tokens", nil, WithHeader("Content-Type", "application/json"), WithBasicAuth(email, password))
 	res := hc.do(t, req)
-	require.Zero(t, res.ContentLength)
+	require.Zero(t, res.ContentLength, "should be zero since no tokens should only be returned within cookies")
```
teleivo, line 179: This sentence is hard to read, do you want to say?

should be zero since tokens should only be returned within cookies

pkg/inttest/http.go
```diff
@@ -185,6 +185,7 @@ func (hc *HTTPClient) SignIn(t *testing.T, email, password string) (*http.Cookie
 	require.Equal(t, http.StatusCreated, res.StatusCode, errMsg+": HTTP status mismatch")
 
 	body, err := io.ReadAll(res.Body)
+	require.NoError(t, err)
```
teleivo, line 188: please add a message like we do in the other requires in the helper `errMsg+"..."`
try to make your tests fail, look at the error and see how you can improve it

radnov (approved):

radnov:
My bad, shouldn't have approved with pending comments and auto-merged enabled.

PR #848, opened 2024-07-23
chore: log name and group when destroying an instance due to TTL expi…

tonsV2:
…ring

### DIFF
--- a/pkg/instance/ttlDestroyConsumer.go
+++ b/pkg/instance/ttlDestroyConsumer.go
@@ -4,6 +4,8 @@ import (
 	"encoding/json"
 	"log/slog"
 
+	"github.com/dhis2-sre/im-manager/pkg/model"
+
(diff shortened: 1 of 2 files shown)


pkg/instance/ttlDestroyConsumer.go
```diff
@@ -43,23 +46,28 @@ func (c *ttlDestroyConsumer) Consume() error {
 			return
 		}
 
-		err := c.instanceDeleter.Delete(payload.ID)
+		instance, err := c.instanceService.FindDeploymentInstanceById(payload.ID)
```
teleivo, line 49: Why do we search it first? Can't we just delete it and log the error which in case it could not be found will say it did not exist 🤔

radnov (approved):

teleivo (approved):

teleivo:
please put the jira issue into the PR description, that should make the github PR discoverable from jira (at least thats what we do in dhis2-core)

PR #857, opened 2024-08-13
feat: batch deploy wrapper and config

Philip-Larsen-Donnelly:
This is a wrapper to deploy-dhis2.sh script. The config file allows setting some global parameters, and defining sets of deployments that will be launched together.

Requires yq 

Docs here: https://github.com/dhis2-sre/im-doc/blob/master/api/README.md#deploy-monolith-instances-in-a-pre-defined-batch

### DIFF
--- a/scripts/instances/deploy-batch.sh
+++ b/scripts/instances/deploy-batch.sh
@@ -1,0 +1,111 @@
+#!/bin/bash
+
+# Ensure yq is installed
+if ! command -v yq &> /dev/null
+then
+    echo "yq could not be found. Please install yq to proceed."
+    exit
+fi
+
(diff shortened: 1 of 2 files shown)


radnov (commented):
Before looking at the changes, I should say that we do have a [bulk deploy script]( https://github.com/dhis2-sre/im-manager/blob/d3d19f63e3130d1775abdc8b271ba45fc4ec4161/scripts/instances/bulk-dhis2-deploy.sh) in the repo already. It is simpler and much more limited than this, but it's still worth discussing if we should replace it with the new one or have the new one in a different repository. 🤔

@tonsV2 any thoughts on this?

PR #884, opened 2024-09-06
docs: add local development environment section to README

radnov:
The intent is to have a minimal step-by-step guide for starting a local development environment (albeit not fully functional).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,28 @@
-# TODO
+# Getting Started
 
-* Write readme
+## Start local development environment with Docker
+
+1. Start by copying the `.env.example` file into a new `.env` file:
+```
+cp .env.example .env
+```
+2. Create a private key:
+```
+make keys
+```
+3. Copy the private key contents, with actual newlines replaced by "\n", into the `PRIVATE_KEY` environment variable within the `.env` file:
+*This should work on macOS to copy the key contents*
+```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,6 +1,32 @@
...
+
+1. Start by copying the `.env.example` file into a new `.env` file:
+```
+cp .env.example .env
+```
+2. Create a private key:
+```
+make keys
+```
+3. Copy the private key contents into the `PRIVATE_KEY` environment variable within the `.env` file:
```
tonsV2, line 17: Maybe describe what needs to be done? Replace actual newlines with \n

tonsV2: Maybe we should even let the make keys command output: PRIVATE_KEY="..." ?

README.md
```diff
@@ -1,6 +1,32 @@
...
+4. Initialize the environment and install dev dependencies:
+```
+make init
+```
+5. Start a development environment:
+```
+make dev
+```
 
 # Add a group
```
tonsV2, line 31: This now seems a bit unrelated. Should we just get rid of it?

tonsV2 (approved):

README.md
radnov: >Maybe we should even let the make keys command output: PRIVATE_KEY="..." ?
I didn't do that because dumping the private key contents on the screen didn't seem like a good practice and copy/pasting from the terminal could also lead to unexpected newlines.

PR #885, opened 2024-09-06
feat: public instances

tonsV2:
This PR changes the way we implement public instances/deployments. Specifically it moves the `public` property from the deployment down to the instance.
For the sake of keeping the UI simple it has been decided that we don't change it as of now but when deploying we simply only set the public property on the dhis2-core instance.

Another breaking change is the introduction of the `Description` property on the `Group` model. When creating a group this property needs to be supplied and once again the frontend needs to be updated.

To sum up, once this PR is merged, two tasks needs to be implemented on the frontend
1. If a deployment is marked as public, the field should from now on be set on the dhis2-core instance and not on the actual deployment
2. A description form field needs to be added to the create group page

A short coming of this PR is that the filter is hard coded. Specifically the instance names have to begin with "dev", "nightly" or "stable". This isn't ideal but it suites the current use case for retrieving the play instances.

https://tons.api.im.dhis2.org/instances/public

NOTE: Delete public column on deployments after deploy


radnov (approved):
LGTM. Filtering the instances to return based on their name isn't ideal, but should do fo now 👍

PR #901, opened 2024-09-25
feat: add option for enabling query logging to dhis2 stack

radnov:
https://dhis2.atlassian.net/browse/DEVOPS-456
https://github.com/dhis2-sre/im-web-client/pull/598


tonsV2 (approved):

teleivo (approved):
Did we think of a way to let users manage their `dhis.conf`?

I can already see many more requests to access properties from

https://github.com/dhis2/dhis2-core/blob/f5b4e5c95740fbec08085b4b22ddd2f916878003/dhis-2/dhis-support/dhis-support-external/src/main/java/org/hisp/dhis/external/conf/ConfigurationKey.java

radnov:
> Did we think of a way to let users manage their `dhis.conf`?
> 
> I can already see many more requests to access properties from
> 
> https://github.com/dhis2/dhis2-core/blob/f5b4e5c95740fbec08085b4b22ddd2f916878003/dhis-2/dhis-support/dhis-support-external/src/main/java/org/hisp/dhis/external/conf/ConfigurationKey.java

Not really, no. We've been thinking of just using a text field and pass the contents as a blob, so that users can set any configuration properties they want, but honestly we haven't had too much request for other options so far. But it's definitely something we'd like to open up a bit more.

In general, what other config properties you think are commonly used and should be included?

teleivo:
> > Did we think of a way to let users manage their `dhis.conf`?
> > I can already see many more requests to access properties from
> > https://github.com/dhis2/dhis2-core/blob/f5b4e5c95740fbec08085b4b22ddd2f916878003/dhis-2/dhis-support/dhis-support-external/src/main/java/org/hisp/dhis/external/conf/ConfigurationKey.java
> 
> Not really, no. We've been thinking of just using a text field and pass the contents as a blob, so that users can set any configuration properties they want, but honestly we haven't had too much request for other options so far. But it's definitely something we'd like to open up a bit more.
> 

We would need to make sure they cannot override configs the IM controls set like the db ones, redis, ...

> In general, what other config properties you think are commonly used and should be included?

I don't know but I see a similar issue with DHIS2 conf in general. We use some libraries that expert users would want to have full control over. For those we need to add yet another property anytime someone wants access to one.

PR #906, opened 2024-09-29
feat: enable tls server side and require it client side

tonsV2:


radnov (approved):
Nice, well done! 👍

pkg/inttest/db.go
```diff
@@ -31,6 +31,7 @@ func SetupDB(t *testing.T) *gorm.DB {
 	db, err := storage.NewDatabase(logger, config.Postgresql{
 		Host:         container.Host,
 		Port:         container.DefaultPort(),
+		SslMode:      "disable",
```
teleivo, line 34: Looks like the `mode` is always `require` except for tests. Is this tested? Making exceptions for tests means our code in tests behaves differently than in prod. If we were able to use the same setup during tests we would not have to add that config at all and be more confident that it works and keeps working.

tonsV2: The issue is that Gnomock doesn't support any ssl mode but disable. Do you have any idea about how we could test this?

teleivo: Add it to gnomock or use testcontainers are the options I know of

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/dhis2-sre/im-manager
 
-go 1.22.1
+go 1.23.0
```
teleivo, line 3: 1.23.2 is out https://go.dev/doc/install

radnov (approved):

PR #926, opened 2024-10-14
chore: remove use of log.Fatal DEVOPS-394

teleivo:
* remove use of log.Fatal so we only exit from main
* remove unused env vars: database manager no longer exists and is in fact the instance manager
  - INSTANCE_SERVICE_BASE_PATH
  - DATABASE_MANAGER_SERVICE_HOST
  - DATABASE_MANAGER_SERVICE_BASE_PATH
* remove `config.Config`: we already started to pull out option like structs we can pass to individual components like `config.Postgresql`.

Next step is to move option like structs like `config.Postgresql` into the actual packages where they are used. Just like a 3rd party dependency that provides options for a `New` function.


teleivo:
close in favor of https://github.com/dhis2-sre/im-manager/pull/927

PR #927, opened 2024-10-15
chore: remove use of log.Fatal DEVOPS-394

teleivo:
* remove use of log.Fatal so we only exit from main
* remove unused env vars: database manager no longer exists and is in fact the instance manager
  - INSTANCE_SERVICE_BASE_PATH
  - DATABASE_MANAGER_SERVICE_HOST
  - DATABASE_MANAGER_SERVICE_BASE_PATH
* remove `config.Config`: we already started to pull out option like structs we can pass to individual components like `config.Postgresql`.
* extract smaller functions to setup components in `main`
  * for this to work we need access to exported types. I therefore exported some of our `Service`s. We should prefer to depend on concrete types of ours unless we want to isolate us during testing or we have multiple implementations. Only then create an interface.


radnov (commented):
Could you elablorate on removing the `config` package? I don't completely understand what is the reason for removing it and spreading the configuration in `main` and other places. To me, having configuration package with all the needed options seems convenient, but maybe it's a bit of an anti-pattern.

Otherwise I do like how you restructured main, now it looks quite less messy with having everything in its own helper function.

teleivo:
> I don't completely understand what is the reason for removing it and spreading the configuration in `main` and other places.

The configuration is actually in one place = `main`. `main` is responsible for instantiating clients, services, repositories, ... and wiring them.

You can see that a lot of the types we created in `config` were not actually needed so they can be removed. Go can easily show us unused parameters and variables. It can also show us unused fields but only if we actively look for them. That is another benefit.

I assume that what you refer to as `spreading in main and other places` is for example

```go
type PostgresqlConfig struct {
	Host         string
	Port         int
	Username     string
	Password     string
	DatabaseName string
}

func NewDatabase(logger *slog.Logger, c PostgresqlConfig) (*gorm.DB, error) {
```

which was previously in `config.Postgresql`.

`storage.NewDatabase` is in charge of its API. This is why such types should be declared in the package itself (look at any lib we are using or the stdlib). There are different patterns for options (plain parameters, structs, [functional options](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis), ...). We can rename these structs if there is a better word then config.

I created 

```go
type DockerHubConfig struct {
	Username string
	Password string
}
```

because if you having parameters with the same type can be risky/make the API hard to use. This way you won't confuse setting the username vs password as easily. See 
https://dave.cheney.net/practical-go/presentations/qcon-china.html#_be_wary_of_functions_which_take_several_parameters_of_the_same_type

radnov (approved):

cmd/serve/main.go
```diff
@@ -239,11 +223,377 @@ func run() (err error) {
...
+		return nil, err
+	}
+	name, err := requireEnv("DATABASE_NAME")
+	if err != nil {
+		return nil, err
+	}
+
+	db, err := storage.NewDatabase(
+		logger,
+		storage.PostgresqlConfig{Host: host, Port: port, Username: username, Password: password, DatabaseName: name},
```
tonsV2, line 250: The variables already have sane names so there's no reason to list the struct keys

```suggestion
		storage.PostgresqlConfig{host, port, username, password, name},
```

pkg/storage/postgresql.go
```diff
@@ -5,14 +5,21 @@ import (
...
-func NewDatabase(logger *slog.Logger, c config.Postgresql) (*gorm.DB, error) {
+type PostgresqlConfig struct {
+	Host         string
+	Port         int
+	Username     string
+	Password     string
+	DatabaseName string
+}
+
+func NewDatabase(logger *slog.Logger, c PostgresqlConfig) (*gorm.DB, error) {
```
tonsV2, line 22: Not this line, but on line 34-38 we're assigning each property from the struct to variables of the same name. How about skipping that and just used the struct?

skaffold.yaml
tonsV2, line 1: Why is the formatting changed here? My IDE disagrees with this :slightly_smiling_face:

tonsV2 (changes requested):

skaffold.yaml
teleivo: https://github.com/redhat-developer/yaml-language-server did that. is there any style you want?

tonsV2 (approved):

PR #928, opened 2024-10-16
chore: pass context to Gorm DEVOPS-397

teleivo:
**Note**: please don't ask again about pointer vs non-pointer. This is not in scope for this change!

> Package context defines the Context type, which carries deadlines, cancellation signals, and other request-scoped values across API boundaries and between processes.
https://pkg.go.dev/context#pkg-overview

This means [context.Context](https://pkg.go.dev/context#Context) has two capabilities, it carries

1. request-scoped values
2. deadlines, cancellation signals

across API boundaries and between processes.

## Previous work

https://github.com/dhis2-sre/im-manager/pull/739 started our use of `context.Context`'s 1. capability. We generate and log an HTTP request id and the user id.

In https://github.com/dhis2-sre/im-manager/pull/741 we setup Gorm so it would use `context.Context`'s 1. capability. Gorm will log slow/failed queries with attributes like the user id we set in the `context.Context` due to our [slog.Handler](https://pkg.go.dev/log/slog#Handler) [ContextHandler](https://github.com/dhis2-sre/im-manager/blob/DEVOPS-397/internal/log/handler.go). [ContextHandler](https://github.com/dhis2-sre/im-manager/blob/DEVOPS-397/internal/log/handler.go) is setup in our `main`. So any context aware log will get the log attributes with values we set in the context.

## Context Cancellation

One important thing to remember is that [context.Context](https://pkg.go.dev/context#Context) also carries deadlines and cancellation signals.

When reading and writing code you need to remember in what scope that particular action is.

* our `main` creates internal users and groups
* our TTL feature is triggered by a message to RabbitMQ
* most of our code is triggered by an HTTP call made from our users

This means that most of our code will be within the "context" of an HTTP request. The `ctx` will then be request-scoped [http.Request.Context](https://pkg.go.dev/net/http#Request.Context). This `ctx` will have the user id (on non-public routes) and a request id. It will also be cancelled if the user cancels the request.

When it comes to Gorm

> Setting a timeout on the context passed to db.WithContext can control the duration of long-running queries. This is crucial for maintaining performance and avoiding resource lock-ups in database interactions.
https://gorm.io/docs/context.html#Context-Timeout

So far we are not setting any timeouts or deadlines. We have everything in place to do so now. We do need to be mindful of request cancellations by the client though. For now, I only used `ctx` for its values when it comes to DB create, update and delete operations (by using [context.WithoutCancel](https://pkg.go.dev/context#WithoutCancel)). This is to avoid any DB rollbacks when a request is cancelled. We would need to look at each case individually. Any DB read will be cancelled if the user cancels the request. This spares IM and our DB some work.

## Changes

* pass `context.Context` to Gorm so it will log slow/failed queries with attributes populated from the context
* let components within a package depend on concrete types instead of interfaces
  * services depend on their concrete repositories
  * handlers depend on their concrete services
* renamed our log attribute key `id` (for the HTTP request id) to `correlationId`. This aligns logs generated from HTTP and RabbitMQ TTL messages.

For example the user handler depends on the user service which depends on the user repository. We should only create interfaces (on the consumer side) if we have multiple implementations or want to isolate for testing or from 3rd party dependencies. You can see test isolation in our integration tests against lets say the user/group services in database or instance handler tests. Our integration tests mostly test one package/handler at a time. Here we isolate as we do not want to re-test these other packages which have tests on their own.

Section `Issues with Gins Context` explains why I think we should stay away from `gin.Context` even though it does implement the [context.Context](https://pkg.go.dev/context#Context) interface. `gin.Context` should only be used in our HTTP handlers and middleware for anything other than `context.Context` capabilities. Whenever a `context.Context` is required in an HTTP context we should rely on the [Request.Context](https://pkg.go.dev/net/http#Request.Context).

## Issues with Gins Context

When moving
* to slog and adding attributes I already hit some odd `gin.Context` issues
* then again when working on SSE with respect to context cancellations

and now again

1. Data race in Gin related to its context 😩 This task gives me flashbacks. We are seeing data races as described in https://github.com/gin-gonic/gin/issues/3931.
2. As https://github.com/gin-gonic/gin/issues/3931#issuecomment-2117737962 and our tests suggest its looks as if its still necessary to pass `c.Request.Context()`. Issue with that is that you have to remember to do that described in https://github.com/gin-gonic/gin/issues/2845. [context.Context](https://pkg.go.dev/context#Context) is an interface so it is valid to pass the `gin.Context`.
3. https://github.com/gin-gonic/gin/issues/3554 is another one showing that we should likely stay away from using `gin.Context` as `context.Context`. We have at least one goroutine that outlives its request. We should not pass the request scoped context but create a dedicated one. Gins copy method does not seem to be usable without surprises.
4. https://github.com/gin-gonic/gin/pull/2029
5. ...

Let the [forbidigo](https://github.com/ashanbrown/forbidigo) linter ensure we use the [Request.Context](https://pkg.go.dev/net/http#Request.Context) instead of the `gin.Context` for values and don't call any of the [context.Context](https://pkg.go.dev/context#Context) interface methods on it.

```sh
pkg/user/handler.go:80:2: use of `c.Set` forbidden because "Always use https://pkg.go.dev/net/http#Request.Context instead of gin.Context" (forbidigo)
        c.Set("foo", 1)

pkg/user/handler.go:81:2: use of `c.Get` forbidden because "Always use https://pkg.go.dev/net/http#Request.Context instead of gin.Context" (forbidigo)
        c.Get("foo")

pkg/user/handler.go:80:6: use of `c.Err` forbidden because "Always use https://pkg.go.dev/net/http#Request.Context instead of gin.Context" (forbidigo)
        _ = c.Err()
```

I don't know of a good way that ensures we always pass the [Request.Context](https://pkg.go.dev/net/http#Request.Context) instead of the `gin.Context` into functions outside of handlers and middleware.

## Manual tests

* sign up
* sign in
* find user that does not exist
* upload databases
* download databases
* list databases
* saving a database
* deploy DHIS2 with TTL
* destroy via inspector due to TTL
* destroy

## Sample logs

How we log with Gorm can be configured here

https://github.com/dhis2-sre/im-manager/blob/1b5136cace379c6701c840eef194996d2eca5b2e/pkg/storage/postgresql.go#L16-L20

Finding a user that does not exist

    ./findUserById.sh 100

The first log is from Gorm the second from our Gin middleware

```json
{
  "time": "2024-10-19T05:19:39.612801689Z",
  "level": "ERROR",
  "msg": "record not found",
  "error": "record not found",
  "query": "SELECT * FROM \"users\" WHERE \"users\".\"id\" = 100 ORDER BY \"users\".\"id\" LIMIT 1",
  "duration": 552112,
  "rows": 0,
  "file": "/src/pkg/user/repository.go:113",
  "correlationId": "91f084af-e256-4a85-84d4-be50bfcf6510",
  "user": 3
}
{
  "time": "2024-10-19T05:19:39.613052745Z",
  "level": "WARN",
  "msg": "Processed HTTP request",
  "correlationId": "91f084af-e256-4a85-84d4-be50bfcf6510",
  "error": "Error #01: failed to find user with id 100\n",
  "request": {
    "time": "2024-10-19T05:19:39.611847197Z",
    "method": "GET",
    "path": "/users/100",
    "route": "/users/:id",
    "query": "",
    "params": {
      "id": "100"
    },
    "host": "devops-397.api.im.dhis2.org",
    "userAgent": "HTTPie/3.2.2"
  },
  "response": {
    "time": "2024-10-19T05:19:39.613011256Z",
    "latency": 1163893,
    "status": 404
  },
  "user": 3
}
```

Uploading a DB:

You can see that we print the progress to stdout which messes up JSON processing of our logs. It is also tons of data that is not useful. Do we need this?

```sh
{
  "time": "2024-10-19T05:22:38.704518131Z",
  "level": "INFO",
  "msg": "Uploading",
  "bucket": "xxx",
  "key": "whoami/40.5.0",
  "correlationId": "9bff80a2-1082-49ed-87c2-391c07284222",
  "user": 3
}
im-databases-feature/whoami/40.5.0 - total read:32768   progress:0%im-databases-feature/whoami/40.5.0 - total read:65536        progress:0%im-databases-feature/whoami/40.5.0 - total read:98304      progress:0%im-databases-feature/whoami/40.5.0 - total read:131072
{
  "time": "2024-10-19T05:22:39.785049734Z",
  "level": "INFO",
  "msg": "Processed HTTP request",
  "correlationId": "9bff80a2-1082-49ed-87c2-391c07284222",
  "request": {
    "time": "2024-10-19T05:22:38.500347106Z",
    "method": "POST",
    "path": "/databases",
    "route": "/databases",
    "query": "",
    "params": {},
    "host": "DEVOPS-397.api.im.dhis2.org",
    "userAgent": "curl/7.88.1"
  },
  "response": {
    "time": "2024-10-19T05:22:39.785033197Z",
    "latency": 1284686044,
    "status": 201
  },
  "user": 3
}
```

./saveAs.sh 19 foo custom

You can see again that we print the progress to stdout which messes up JSON processing of our logs. It is also tons of data that is not useful. Do we need this?

```json
{
  "time": "2024-10-19T12:50:08.160074139Z",
  "level": "INFO",
  "msg": "Processed HTTP request",
  "correlationId": "69126be1-577c-4cfd-9f1a-481098119b0f",
  "request": {
    "time": "2024-10-19T12:50:08.148019705Z",
    "method": "POST",
    "path": "/databases/save-as/19",
    "route": "/databases/save-as/:instanceId",
    "query": "",
    "params": {
      "instanceId": "19"
    },
    "host": "devops-397.api.im.dhis2.org",
    "userAgent": "HTTPie/3.2.2"
  },
  "response": {
    "time": "2024-10-19T12:50:08.160054109Z",
    "latency": 12034417,
    "status": 201
  },
  "user": 3
}
pg_dump: last built-in OID is 16383
pg_dump: reading extensions
pg_dump: identifying extension members
...
progress:99%im-databases-feature/whoami/foo - total read:86114304       progress:99%im-databases-feature/whoami/foo - total read:86147072       progress:99%im-databases-feature/whoami/foo - total read:86179840progress:99%im-databases-feature/whoami/foo - total read:86201923       progress:100%
{
  "time": "2024-10-19T12:50:24.098223809Z",
  "level": "INFO",
  "msg": "Save an instances database as",
  "groupName": "whoami",
  "databaseName": "foo",
  "instanceName": "test-ivo-5",
  "correlationId": "69126be1-577c-4cfd-9f1a-481098119b0f",
  "user": 3
}
```

## Helpful Articles

* https://pkg.go.dev/context
* https://gorm.io/docs/context.html
* https://fideloper.com/golang-context-http-middleware
* https://www.alexedwards.net/blog/how-to-manage-database-timeouts-and-cancellations-in-go
* https://pkg.go.dev/log/slog#Handler


radnov (approved):

pkg/database/database_integration_test.go
```diff
@@ -246,16 +247,16 @@ type groupService struct {
...
 		Name: gs.groupName,
 	}, nil
 }
 
 type instanceService struct{}
 
-func (is instanceService) FindDecryptedDeploymentInstanceById(id uint) (*model.DeploymentInstance, error) {
-	//TODO implement me
+func (is instanceService) FindDecryptedDeploymentInstanceById(ctx context.Context, id uint) (*model.DeploymentInstance, error) {
+	// TODO implement me
```
tonsV2, line 259: We shouldn't have TODO's int he code that really isn't a todo

tonsV2 (approved):

PR #967, opened 2024-12-12
feat: instance details

tonsV2:
The goal of this PR is to provide two new endpoints
* /instances/:id/details
* /instances/:id/decrypted-details

They can be tested using the following user scripts
* scripts/instances/findDeploymentInstanceById.sh
* scripts/instances/findDeploymentInstanceDecryptedById.sh


radnov (approved):
LGTM 👍

PR #970, opened 2024-12-19
feat: only encrypt sensitive parameters

tonsV2:


pkg/stack/stack.go
```diff
@@ -186,9 +186,9 @@ var DHIS2Core = model.Stack{
 		"IMAGE_PULL_POLICY":               {Priority: 3, DisplayName: "Image Pull Policy", DefaultValue: &dhis2CoreDefaults.imagePullPolicy, Validator: imagePullPolicy},
 		"STORAGE_TYPE":                    {Priority: 4, DisplayName: "Storage type", DefaultValue: &dhis2CoreDefaults.storageType, Validator: storage},
 		"S3_BUCKET":                       {Priority: 5, DisplayName: "S3 bucket", DefaultValue: &dhis2CoreDefaults.s3Bucket},
```
radnov, line 188: S3 bucket names in general could also be considered sensitive in some contexts, doesn't really apply to us but maybe we can make it sensitive, as well.

radnov (approved):
Looks good to me. One minor suggestion added. 👍

PR #981, opened 2025-01-31
feat: restore FS

tonsV2:
The goal of this PR is to implement restoring of FS.

It's a bit of a hack, it's using a side car container to seed MinIO which means the container lingers around as long as the main container is up. Ideally this should be done using an init container but that would require starting MinIO within the init container. While this is more work it might not be that difficult.


radnov (approved):
Added two comments, otherwise looks good, nice work 👍

PR #1003, opened 2025-02-13
Remove decrypted instance endpoint

tonsV2:


pkg/instance/handler.go
```diff
@@ -91,9 +95,40 @@ func (h Handler) DeployDeployment(c *gin.Context) {
...
+
+func (h Handler) stripInstanceSensitiveParameterValues(instance *model.DeploymentInstance) error {
+	stack, err := h.stackService.Find(instance.StackName)
+	if err != nil {
+		return err
+	}
+
+	for index, parameter := range instance.Parameters {
+		if stack.Parameters[parameter.ParameterName].Sensitive {
+			parameter.Value = "!!!redacted!!!"
```
radnov, line 125: This looks a bit weird to me. If there's no better way of hiding the value, maybe we can use `"***"` or `"sensitive"` 🤷

radnov (approved):
Looks good 👍

PR #1014, opened 2025-02-24
Feat/opentelemetry

tonsV2:
Screenshot From 2025-02-24 22-37-16


cmd/serve/main.go
```diff
@@ -214,6 +226,8 @@ func run() (err error) {
 		return err
 	}
 
+	r.Use(otelgin.Middleware("my-app")) // Attach OpenTelemetry middleware
```
radnov, line 229: Maybe name the middleware `im` or something similar, instead of `my-app`

radnov (approved):

PR #1018, opened 2025-02-25
chore: generic backup source and concrete MinIO implementation

tonsV2:
The goal of this PR is to introduce a generic backup source and refactor the MinIO specific backup logic in to a separate struct.
The next step would be to implement backup functionality for backing up a DHIS2 instance which is using the filesystem for storage rather than MinIO or AWS S3.


radnov (approved):

radnov:
Just a nitpick, but these changes seem more like a refactor than a chore.

PR #1089, opened 2025-05-15
chore: share bulk operation scripts

Philip-Larsen-Donnelly:


radnov (commented):
LGTM, though I haven't tested the scripts.

One thing to note is that there's a mix of Httpie and Curl used in the scripts. If possible I'd go with Httpie everywhere, since it's already the "convention" in other user scripts.

Philip-Larsen-Donnelly:
> LGTM, though I haven't tested the scripts.
> 
> One thing to note is that there's a mix of Httpie and Curl used in the scripts. If possible I'd go with Httpie everywhere, since it's already the "convention" in other user scripts.

OK. I've replaced curl with Httpie.

radnov (approved):
Ship it! 🚀

PR #1126, opened 2025-06-19
fix: remote status

tonsV2:


Philip-Larsen-Donnelly (approved):
Looks reasonable to me! :+1:

PR #1139, opened 2025-07-03
fix: save

tonsV2:
In advance I'd like to apologize for this PR

* Include Locks when listing Databases
* Fix path when unlocking a database
* Use Group.Namespace rather than Instance.GroupName when interacting with the cluster
* Fix save by unlocking before deleting the original database

Save seems to work but I'm getting the below errors

    {"time":"2025-07-04T00:03:06.03324424Z","level":"ERROR","msg":"Failed to remove temp file","error":"close /mnt/data/e4526bab-8d89-4550-a040-750e6e6ab2ab.dump: file already closed","correlationId":"4c39fa85-d8fb-4e75-a0cb-1ad4732b7f47","user":12}

### DIFF
--- a/scripts/databases/unlock.sh
+++ b/scripts/databases/unlock.sh
@@ -6,4 +6,4 @@ source ./auth.sh
 
 DATABASE=$1
 
-$HTTP delete "$IM_HOST/databases/$DATABASE/unlock" "Authorization: Bearer $ACCESS_TOKEN"
+$HTTP delete "$IM_HOST/databases/$DATABASE/lock" "Authorization: Bearer $ACCESS_TOKEN"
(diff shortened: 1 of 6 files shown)


scripts/databases/unlock.sh
radnov, line 1: This seems wrong - using the lock endpoint in the unlock script.

radnov (approved):

scripts/databases/unlock.sh
tonsV2: The script was using the wrong path. Here I'm simply fixing the script.

You create a lock by sending a post request to /databases/:id/lock and you unlock by sending a delete request to the same url. Doesn't that make sense? :-)

PR #1375, opened 2026-02-24
test: add SaveAs and BackupService integration tests

radnov:
### New/updated tests

- **TestInstanceHandler/SaveAsDatabase** - Integration test that deploys the dhis2-db stack into k3s and drives the SaveAs HTTP endpoint end-to-end: real Postgres pod, real pg_dump streamed into S3, async DB record update verified, and S3 object content checked.
- **TestBackupServiceIntegration** - Integration test for BackupService with MinIO (testcontainer) and Localstack S3; seeds MinIO, runs PerformBackup, and checks that the S3 tar.gz matches MinIO paths and content (replaces the previously skipped test).


tonsV2 (changes requested):

pkg/database/database_integration_test.go
```diff
@@ -269,6 +271,121 @@ func TestDatabaseHandler(t *testing.T) {
...
+		group := &model.Group{
+			Name:      "packages",
+			Hostname:  "some",
+			Namespace: "ns",
+		}
+		user := &model.User{
+			Email:  "user1@dhis2.org",
+			Groups: []model.Group{*group},
+		}
+		db.Create(user)
```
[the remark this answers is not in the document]
radnov: I was initially running just this test in isolation, so this was needed. I guess we can reuse the setup from the outer test or refactor this into a helper. 🤔

tonsV2 (changes requested):

tonsV2 (changes requested):

pkg/database/database_integration_test.go
```diff
@@ -269,6 +271,121 @@ func TestDatabaseHandler(t *testing.T) {
...
+			Parameters: model.DeploymentInstanceParameters{
+				"DATABASE_ID":       {Value: sourceDBID},
+				"DATABASE_NAME":     {Value: "dhis2"},
+				"DATABASE_USERNAME": {Value: "dhis"},
+				"DATABASE_PASSWORD": {Value: "dhis"},
+			},
+		}
+		db.Create(instance)
+
+		fakePodExecutor := &fakePodExecutor{output: []byte("pg_dump fake output")}
```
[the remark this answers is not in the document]
tonsV2: I think maybe around [here](https://github.com/dhis2-sre/im-manager/blob/d0b1715ac3d2e0f6c82e6f9b43e36779be13f3ed/pkg/instance/instance_integration_test.go#L170) would be good :grin:

pkg/database/database_integration_test.go
```diff
@@ -20,30 +20,25 @@ import (
...
-	s3Bucket := "database-bucket"
-	err := os.Mkdir(s3Dir+"/"+s3Bucket, 0o755)
-	require.NoError(t, err, "failed to create S3 output bucket")
-	logger := slog.New(slog.NewTextHandler(os.Stdout, nil))
-	s3 := inttest.SetupS3(t, s3Dir)
-	uploader := manager.NewUploader(s3.Client)
-	s3Client := storage.NewS3Client(logger, s3.Client, uploader)
+	env := setupDatabaseHandlerEnv(t)
+	db := env.db
+	var err error
```
tonsV2: Why? 🙂

pkg/database/database_integration_test.go
```diff
@@ -20,30 +20,25 @@ import (
...
-
-	s3Dir := t.TempDir()
-	s3Bucket := "database-bucket"
-	err := os.Mkdir(s3Dir+"/"+s3Bucket, 0o755)
-	require.NoError(t, err, "failed to create S3 output bucket")
-	logger := slog.New(slog.NewTextHandler(os.Stdout, nil))
-	s3 := inttest.SetupS3(t, s3Dir)
-	uploader := manager.NewUploader(s3.Client)
-	s3Client := storage.NewS3Client(logger, s3.Client, uploader)
+	env := setupDatabaseHandlerEnv(t)
```
tonsV2: Could we come up with a slightly better name for this variable?

pkg/instance/backupService_test.go
```diff
@@ -1,61 +1,192 @@
...
 
+func TestBackupService(t *testing.T) {
+	t.Parallel()
+
+	testFiles := map[string][]byte{
+		"apps/app1/manifest.json": []byte(`{"name":"app1"}`),
+		"userAvatar/uid1":         []byte("avatar-content"),
+	}
+
+	source := &memoryBackupSource{objects: testFiles}
```
tonsV2: Why are we doing this in memory now rather than using the actual services?

tonsV2 (changes requested):

pkg/instance/backupService_test.go
radnov: I think we can delete the `TestBackupService`, following the same logic as the SaveAs test - just have an integration test, no need for a duplicated unit test.

pkg/database/database_integration_test.go
radnov: Will inline the setup as it was before, we don't need this anymore.

pkg/database/database_integration_test.go
radnov: Will remove this, as it's no longer needed after inlining the setup.

pkg/instance/backupService_test.go
tonsV2: Only thing is that we support backup from multiple sources such as plain volumes and MinIO

pkg/database/test_helpers.go
```diff
@@ -0,0 +1,58 @@
...
+		inttest.WithHeader("Content-Type", writer.FormDataContentType()),
+	}
+
+	if len(headers) > 0 {
+		baseHeaders = append(baseHeaders, headers...)
+	}
+
+	body := client.Put(t, "/databases", &buffer, http.StatusCreated, baseHeaders...)
+
+	var actualDB model.Database
```
tonsV2: s/actualDB/database/ ?

tonsV2 (approved):

PR #1421, opened 2026-03-25
feat: Doris

tonsV2:
# Doris

## TODO

* Docker images... Postgresql.jar
* MinIO logs 500... Probably not a Doris related issue

### Monolith issue

* Status endpoint is not working
* Reset/restart/logs endpoints are not working
*


radnov (changes requested):

tonsV2:
Superseded by the chart-level approach: the dhis2 chart bundles the Doris cluster, so Doris landed as parameters and components of dhis2-v2 in #1703 and #1705 rather than as a stack of its own.

PR #1434, opened 2026-04-01
feat: chap

tonsV2:


cmd/serve/main.go
```diff
@@ -250,9 +250,19 @@ func run() (err error) {
 	event.Routes(r, authentication.TokenAuthentication, eventHandler)
 
 	logger.InfoContext(ctx, "Listening and serving HTTP")
-	if err := r.Run(); err != nil {
+	//	if err := r.Run(); err != nil {
+	//		return fmt.Errorf("failed to start the HTTP server: %v", err)
+	//	}
```
radnov: Should this be deleted?

pkg/integration/ghcr.go
radnov, line 1: No HTTP status checks here compared to the dockerhub equivalent.

pkg/integration/ghcr.go
```diff
@@ -0,0 +1,97 @@
...
+		if t == tag {
+			return nil
+		}
+	}
+
+	return fmt.Errorf("%s/%s:%s not found", organization, repository, tag)
+}
+
+func (g *ghcrClient) GetImages(_ string) ([]string, error) {
+	return nil, nil
```
radnov, line 105: Are there other options here besides silently returning `nil`?

pkg/instance/service.go
```diff
@@ -416,7 +416,12 @@ func (s Service) deployDeploymentInstance(ctx context.Context, token string, ins
 	}
 	*/
 	if err != nil {
-		return err
+		// TODO: This is a hack to detect if the helmfile operation is already in progress.
```
radnov, line 419: Why do we need this?

stacks/chap-core/helmfile.yaml.gotmpl
```diff
@@ -0,0 +1,43 @@
+# consumedParameters: DATABASE_HOSTNAME,DATABASE_NAME,DATABASE_SECRET,REDIS_HOST,REDIS_SECRET
+helmDefaults:
+  deleteWait: true
+  timeout: 420
```
radnov: Do we need this explicit timeout here? 👀

stacks/dhis2-core/helmfile.yaml.gotmpl
```diff
@@ -79,6 +79,18 @@ releases:
...
+          {{- end }}
+
+      {{- if eq (requiredEnv "DEPLOY_CHAP") "true" }}
+      - jobs:
+          installApps:
+            enabled: true
+            apps:
+              - name: Modeling
+                version: latest
+      {{- end }}
```
radnov, line 93: Image

Seems like the `route.remote_servers_allowed=http://*` might be breaking the syntax or it's just the GitHub preview that's broken. 🤔

pkg/instance/service.go
tonsV2: Because that error is reported which cause our endpoint to return an error but the cluster fixes itself afterwards. I'd love to find a better solution! Especially one where the error doesn't occur at all but I've spent more time on this already than I should.

stacks/dhis2-core/helmfile.yaml.gotmpl
tonsV2: It works. Otherwise DHIS2 wouldn't be able to talk to CHAP at all

pkg/integration/ghcr.go
tonsV2: I've added a comment explaining why this works

cmd/serve/main.go
```diff
@@ -253,6 +253,16 @@ func run() (err error) {
 	if err := r.Run(); err != nil {
 		return fmt.Errorf("failed to start the HTTP server: %v", err)
 	}
+	server := &http.Server{
+		Addr:         ":8080",
```
radnov, line 257: Shouldn't this be an env var?

radnov (approved):

PR #1466, opened 2026-04-22
fix: resolve helmfile and helm binaries at startup via PATH

radnov:
## Summary
- Replace hardcoded `/usr/bin/helmfile` and `/usr/bin/helm` with lookup via `exec.LookPath` at service construction, with optional `HELMFILE_BIN` / `HELM_BIN` overrides.
- Fails fast at startup with a clear error if either binary is missing.
- Logs the resolved paths so operators can audit which binaries got picked.

## Why this isn't an extra security risk
The original `#nosec G204` stance still applies. The resolved paths are:
- computed **once at process startup** from trusted sources (env vars + `PATH`),
- stored on the struct and reused on every request,
- never influenced by user input.

At request time, behavior is identical to the old hardcode: `exec.Command` runs an absolute path that nothing user-facing can change. The only expanded attack surface is at startup: an attacker with write access to a directory on `PATH` could drop a malicious `helmfile`. In the container image we ship, `PATH` is baked in at build time and the filesystem is not writable by the service, so this vector requires a host compromise that would already grant equivalent or greater control.

## Why this matters
On Mac helmfile is at `/opt/homebrew/bin/helmfile` or `/usr/local/bin/helmfile`, on Linux/CI it's at `/usr/bin/helmfile`. The old code only worked on the latter, making the integration tests unrunnable locally on MacOS.


pkg/instance/helmfile.go
```diff
@@ -16,20 +16,50 @@ import (
...
 		stackService:   stackService,
 		stackFolder:    stackFolder,
 		classification: classification,
+		helmfileBin:    helmfileBin,
+		helmBin:        helmBin,
+	}, nil
+}
+
+// resolveBinary returns an absolute path to name, resolved once at startup.
+// An explicit override via envVar wins; otherwise PATH is consulted.
```
tonsV2: Why would we ever want to override this?

pkg/instance/helmfile.go
```diff
@@ -67,7 +97,7 @@ func (h helmfileService) executeHelmfileCommand(ctx context.Context, token strin
 		return nil, err
 	}
 
-	cmd := exec.Command("/usr/bin/helmfile", "--helm-binary", "/usr/bin/helm", "-f", stackPath, operation) // #nosec
+	cmd := exec.Command(h.helmfileBin, "--helm-binary", h.helmBin, "-f", stackPath, operation) // #nosec
```
tonsV2: Minor detail but could we please use "binary" rather than "bin" 🙏

tonsV2 (approved):

pkg/instance/helmfile.go
radnov: I thought it won't hurt having the option, but right now we don't really have the use case, so I'll remove it.

PR #1475, opened 2026-04-30
fix(instance, stack): UpdateInstance validation + FindAll response shape [DEVOPS-424]

radnov:
- Fixes `UpdateInstance` rejecting consumed parameters that were previously resolved and persisted. The validator now checks only incoming parameters, not the fully merged instance.
- Fixes `/stacks` (FindAll) returning `parameters` as a map; the OpenAPI `Stacks` response declares an array, and `/stacks/{name}` already returns one. The list endpoint now uses the same conversion helper, so both responses match the spec.
- Reuses `rejectConsumedParameters` across `SaveInstance` and `UpdateInstance` instead of duplicating the loop.


pkg/instance/service.go
```diff
@@ -947,18 +947,28 @@ func (s Service) UpdateInstance(ctx context.Context, token string, deploymentId,
 		instance.Public = *public
 	}
 
+	stack, err := s.stackService.Find(instance.StackName)
+	if err != nil {
+		return nil, err
+	}
+
+	var consumedErrs []error
```
tonsV2: Why not still use the rejectConsumedParameters method? We're wrapping its logic because we're using it in two places.

tonsV2 (approved):

pkg/instance/instance_integration_test.go
```diff
@@ -261,16 +261,19 @@ func TestInstanceHandler(t *testing.T) {
...
+		createDHIS2DBInstance(t, client, deployment.ID, databaseID, tokens.AccessToken)
+		createMinioInstance(t, client, deployment.ID, tokens.AccessToken)
+		deploymentInstance := createDHIS2CoreInstance(t, client, deployment.ID, tokens.AccessToken,
+			WithParameter("IMAGE_TAG", "2.40.0"),
+			WithParameter("ALLOW_SUSPEND", "false"),
 			WithPublic(false))
 
 		updatedInstance := updateInstance(t, client, deploymentInstance, tokens.AccessToken,
-			WithParameter("IMAGE_TAG", "0.7.0"),
+			WithParameter("IMAGE_TAG", "2.40.1"),
```
tonsV2: Should we use a more recent version?

pkg/instance/service.go
```diff
@@ -145,14 +146,14 @@ func (s Service) SaveInstance(ctx context.Context, instance *model.DeploymentIns
...
+func (s Service) rejectConsumedParameters(stackName string, paramNames iter.Seq[string]) error {
+	stack, err := s.stackService.Find(stackName)
 	if err != nil {
 		return err
 	}
 
 	var errs []error
-	for name := range instance.Parameters {
+	for name := range paramNames {
 		if stack.Parameters[name].Consumed {
```
tonsV2, line 157: Depending on the source this might not be that important but this could result in panic if name does not exist in stack.Parameters

tonsV2: Does name come from the client?

pkg/stack/stack_integration_test.go
```diff
@@ -35,19 +34,25 @@ func TestStackHandler(t *testing.T) {
...
 
 	t.Run("GetAllStacks", func(t *testing.T) {
 		t.Parallel()
 
-		var stacks []model.Stack
+		var stacks []stack.Stack
 		client.GetJSON(t, "/stacks", &stacks)
 
 		assert.NotEmpty(t, stacks)
+		for _, s := range stacks {
```
tonsV2, line 51: There has to be a method on assert which does this

tonsV2 (changes requested):

pkg/instance/service.go
radnov: Yes, `name` is supplied by the client in both places.
Should be safe regardless, because map value type is StackParameter (not pointer), so missing keys return the zero value and Consumed is false. 

 If you mean that we should also reject unknown parameter names (not just consumed ones), I think it should be a separate PR. :thinking:

pkg/stack/stack_integration_test.go
radnov: Testify has no `AllSatisfy`/`Every` or something of that nature - `assert.Condition` would drop the per-element failure message. The loop should be the clearest option here, but we can swap to a collect then `assert.Empty` pattern if you think that's better.

pkg/instance/service.go
tonsV2: I stand corrected! :+1:

tonsV2 (approved):

PR #1499, opened 2026-05-13
chore: drop unused build dependencies

tonsV2:
* Removes unused gcc/musl-dev from the Dockerfile build stage (CGO is disabled so they were never needed)
* Removes the broken `go install github.com/direnv/direnv@latest` from `make init`, which resolved to an ancient v2.20.1 due to module resolution issues; install direnv via your OS package manager or arkade instead.


radnov (commented):
PR title/desc seems out of sync with changes - doesn't seem like the downstream PR https://github.com/dhis2-sre/gha-workflows/pull/103 implements parallelisation.

radnov (approved):

PR #1512, opened 2026-05-15
fix: document and make explicit binary verification for all Dockerfile fetches

tonsV2:
Fixes DEVOPS-157.

**State before this PR:** kubectl, helm, helmfile, and aws-iam-authenticator were already pinned to explicit versions with SHA256 checksums verified during the build. reflex was pinned by version tag but the verification mechanism was invisible.

**Change:** Added a comment and explicit `GONOSUMDB=""` to the reflex install step. `go install @version` verifies the module hash against Go's checksum database (sum.golang.org), a tamper-evident transparent log. Setting `GONOSUMDB=""` makes it explicit that this check is not bypassed.

reflex is a dev-only binary (docker-compose hot reload) and is not copied to the final image. The four binaries in the final image are all SHA256-verified.

**Signatures vs checksums — upstream summary:**

| Binary | Verification available |
|--------|----------------------|
| kubectl | SHA256 checksum (used). cosign signature also available since k8s 1.24. |
| helm | SHA256 checksum (used). GPG signature also available via keys.helm.sh. |
| helmfile | SHA256 checksum (used). No GPG/cosign signature available. |
| aws-iam-authenticator | SHA256 checksum (used). No GPG/cosign signature available. |
| reflex | Go module checksum database (sum.golang.org). No release binaries or signatures published. |

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,12 +22,6 @@ ARG AWS_IAM_AUTHENTICATOR_VERSION=0.7.5
 ARG AWS_IAM_AUTHENTICATOR_CHECKSUM_AMD64=aef183b5b92f2cb135107234c7440f43638caa337190190cdd2ad9fd6bc4928e
 ARG AWS_IAM_AUTHENTICATOR_CHECKSUM_ARM64=80ab2b0d5a139e55408c785703e8fef8f4974a73409046b3e13df19b2a780a51
 
-# https://github.com/cespare/reflex/releases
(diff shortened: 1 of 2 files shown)


radnov (commented):
Not sure if this is ready for review, but the PR description and changes (just removing reflex) seem out of sync, or I don't get it.

tonsV2:
Closing as the changes here were superseded by #1514 which already removed reflex from the Dockerfile and docker-compose.

PR #1518, opened 2026-05-18
feat: optionally log SQL queries via DATABASE_LOG_QUERIES

tonsV2:
## Summary

Enables per-query SQL logging via slog-gorm's `WithTraceAll()` when `DATABASE_LOG_QUERIES=true`. Defaults off so prod stays quiet. Slow-query and error logging are unchanged.

## Test plan

- [ ] Set `DATABASE_LOG_QUERIES=true` locally and confirm `SQL query executed [...]` log lines appear with `query`, `duration`, `rows`, `file`, and `correlationId` fields.
- [ ] Leave it unset (or `false`) and confirm only slow-query/error logs are emitted.

### DIFF
--- a/cmd/serve/main.go
+++ b/cmd/serve/main.go
@@ -296,9 +296,11 @@ func newDB(logger *slog.Logger) (*gorm.DB, error) {
 		return nil, err
 	}
 
+	logQueries := os.Getenv("DATABASE_LOG_QUERIES") == "true"
+
 	db, err := storage.NewDatabase(
 		logger,
-		storage.PostgresqlConfig{Host: host, Port: port, Username: username, Password: password, DatabaseName: name},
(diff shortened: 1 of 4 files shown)


cmd/serve/main.go
```diff
@@ -296,9 +296,11 @@ func newDB(logger *slog.Logger) (*gorm.DB, error) {
 		return nil, err
 	}
 
+	logQueries := os.Getenv("DATABASE_LOG_QUERIES") == "true"
```
radnov, line 299: To me this seems like it should be opt-in rather than opt-out.

tonsV2: It is opt-in 😉

radnov: My bad, I misread that. 👍

radnov (approved):

PR #987, opened 2025-02-05
chore: use Argon2id for password hashing

tonsV2:
Replaces scrypt with Argon2id for password hashing and enforces a 24-character minimum on new passwords (matching the existing handler-level validation).

Note: this branch is well behind master and existing scrypt-hashed users will not be able to sign in after this lands without a migration or password reset. The branch needs a rebase before review.

PR #1525, opened 2026-05-19
chore: hash passwords with Argon2id and migrate scrypt hashes on sign-in

tonsV2:
## Summary

- Switch password hashing from scrypt to Argon2id, written in the standard `$argon2id$v=19$m=...,t=...,p=...$salt$hash` encoding.
- Keep scrypt verification alive for legacy hashes so existing users are not locked out.
- On a successful sign-in against a legacy hash, rehash with Argon2id and persist. Failures of the rehash/persist are logged but never fail the sign-in, so the next attempt retries.

Existing scrypt hashes stay valid until each user signs in once. Replaces #987, which switched the algorithm without a migration path.

## Test plan

- [x] `go test ./pkg/user/...` covers Argon2id round-trip, legacy scrypt verification, malformed-hash error paths, and the end-to-end migration through `POST /tokens`.
- [x] CI green.


radnov (commented):
I wonder how would we know when all users have been "migrated" to the new hashing algo? Query the DB once in a while and remove the redundant code once everyone is migrated or?

tonsV2:
Yeah, or just leave it in there.

radnov (approved):

PR #1541, opened 2026-05-22
fix: exclude apps/ from filestore backup to prevent accumulation

tonsV2:
DHIS2 bundles ~30 core apps into MinIO under `apps/` on every startup. These are not user data; DHIS2 regenerates them from the WAR on each boot. Including `apps/` in filestore backups caused the filestore to grow to hundreds of MB across backup/restore cycles: `mc mirror` deposited old app directories alongside fresh ones, DHIS2 logged "Could not find manifest file" errors for any restored directory it did not recognise, then re-bundled all apps under new UIDs, leaving the old directories stranded.

Fix: add `excludePrefixes` support to `MinioBackupSource` and exclude `apps/` from `FilestoreBackup`.

Also fixes two latent bugs in `backupService.go`: pipe writer now calls `CloseWithError` so tar stream errors propagate to the S3 upload goroutine and trigger an abort (previously a clean EOF was sent on error), and `streamToS3WithMultipart` uses a named return so the deferred cleanup actually sees the return error (loop variable shadowing made it always nil). A unit test is added to prove the abort behaviour; the integration test is updated to verify `apps/` is excluded.

Closes DEVOPS-679

### DIFF
--- a/pkg/instance/backupService.go
+++ b/pkg/instance/backupService.go
@@ -80,8 +80,9 @@ func (s *BackupService) PerformBackup(ctx context.Context, s3Bucket, key string)
 	pr, pw := io.Pipe()
 
 	g.Go(func() error {
-		defer pw.Close()
-		return s.createTarGzStream(ctx, pw, stats)
+		err := s.createTarGzStream(ctx, pw, stats)
+		_ = pw.CloseWithError(err)
+		return err
 	})
 
 	g.Go(func() error {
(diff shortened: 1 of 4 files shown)


tonsV2:
Closed on purpose: we need apps/ in the filestore backup. The database dump and the filestore tarball are one snapshot, and DHIS2 keeps installed-app metadata in the database, so a restore that never writes the apps/ keys the restored database references leaves the two halves disagreeing. Excluding apps/ would also drop user-installed custom apps, which are not regenerated from the WAR the way the bundled core apps are.

The accumulation problem this PR describes is fixed on dhis2 core master, so it only affects the versions that predate the backport. The core developers have been told, and DEVOPS-679 is closed. IM is not carrying a workaround for it. If it ever resurfaces, the fix belongs on the restore side, replacing apps/ before mirroring the backup in rather than merging it with whatever core bundled on first boot, and it would need doing on every restore path: the minio seed sidecar, the filesystem seed script and the external S3 restore in IM.

PR #1542, opened 2026-05-22
fix: clear apps/ from MinIO before seeding to prevent accumulation

tonsV2:
DHIS2 re-bundles its core apps into `apps/<name>_<uid>/` in MinIO on every startup. Because `mc mirror` is additive, old app directories from a previous backup accumulate alongside newly-created ones on each restore cycle, causing the filestore to grow unboundedly (reported at 600MB+).

The fix deletes `apps/` from MinIO before running `mc mirror`, so each restore starts clean. User-installed apps are preserved because they are included in the filestore backup.

### DIFF
--- a/stacks/minio/seed-minio.sh
+++ b/stacks/minio/seed-minio.sh
@@ -57,6 +57,7 @@ else
     gunzip -c "$tmp_file" | tar xf - -C "$tmp_dir"
     chmod -R u+rwx,go+rx "$tmp_dir"
 
+    mc rm --recursive --force --quiet local/dhis2/apps/ || true
     mc mirror "$tmp_dir"/ local/dhis2/
 
     echo "Seeded from $FILESTORE_DOWNLOAD_URL" | mc pipe $seed_file


tonsV2:
Closing for now. Investigation revealed the actual root cause is in DHIS2 itself: orphaned app directories without manifest.webapp accumulate when the DHIS2 pod gets SIGTERM'd mid-cleanup (interrupted between deleting the old manifest and the rest of the old app's files). The orphans are then invisible to DHIS2's discoverInstalledApps scan, so removeOtherAppsWithSameKey can never clean them up.

dev-7 shows 37 metadata-management copies and 26 maps copies, all manifest-less, accounting for ~1.5GB of the 1.58GB filestore. The pod is currently crash-looping with liveness/startup probe failures, which is the interruption source.

Clearing apps/ before mc mirror doesn't help, because the orphans are inside the backup itself. The right fix lives in DHIS2 (re-order deleteApp to remove manifest last, or add an orphan sweep on startup), not in im-manager's seed script.

PR #1591, opened 2026-07-01
fix: surface MinIO listing errors from backup source

tonsV2:
MinioBackupSource.List started an errgroup goroutine but discarded the group without ever calling Wait, so a ListObjects failure mid-stream was silently dropped: the channel just closed and the backup proceeded as if the listing had completed, producing a backup that is missing objects with no error reported. BackupService already checks BackupObject.Err on every item it reads from the channel, so the fix sends the listing error through the channel via that existing field instead of returning it from an unawaited goroutine. Adds a unit test with a fake MinioClient that fails mid-listing.

### DIFF
--- a/pkg/instance/minioBackupSource.go
+++ b/pkg/instance/minioBackupSource.go
@@ -7,7 +7,6 @@ import (
 	"log/slog"
 
 	"github.com/minio/minio-go/v7"
-	"golang.org/x/sync/errgroup"
 )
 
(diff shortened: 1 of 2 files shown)


tonsV2:
Closing as superseded by master. The filestore backup/restore feature (49be9329) independently landed the same fix: MinioBackupSource.List now surfaces listing errors through BackupObject.Err (which the consumer already checks), and master carries an equivalent test, TestMinioBackupSourceListPropagatesError in filestore_test.go, plus a restore-marker skip test. Merging this would only duplicate that coverage, so there's nothing left to bring in.

PR #1594, opened 2026-07-14
refactor: introduce a deployment service to remove the instance/database circular dependencies

tonsV2:
The instance and database services each held a reference back to the other: seed download links were wired through an externalDownloadCreator interface, a nil field and a SetExternalDownloads setter called from main after both services existed, and the filestore save was piggybacked onto the database save via a done callback chain and a FilestoreBackuper interface. A new pkg/deployment service now sits above both and owns every cross-service operation: DeployDeployment, UpdateDeployment, Reset, UpdateInstance and the save database then save filestore workflow. Both leaf services are independent, wiring is plain constructor injection, database.Service exposes blocking primitives (CreateDatabase, Dump, EnsureLocked, SaveLocked) instead of running its own goroutine, and filestore-backup events moved to pkg/deployment with an unchanged wire format. Two deliberate behavior changes: pg_dump configuration errors surface as error events instead of failing the HTTP request, and the filestore backup no longer runs after a failed dump or finalize. Stacked on #1562 and includes its cross-instance DATABASE_ID resolution ported into the deployment service.


pkg/database/service.go
```diff
@@ -433,95 +425,113 @@ func (s Service) FindExternalDownload(ctx context.Context, uuid uuid.UUID) (*mod
...
+	saved, err := s.Dump(ctx, database.UserID, created, instance, stack, format)
+	if err != nil {
+		return nil, err
+	}
 
-		err = s.repository.Unlock(ctx, database.ID)
-		if err != nil {
-			s.logError(ctx, fmt.Errorf("unlock database failed: %v", err))
-		}
+	return s.finalizeSave(ctx, database, saved, wasLocked)
```
radnov: `EnsureLocked` locks the database synchronously when it wasn't already locked, but the only unlock lives in `finalizeSave`'s defer. If `CreateDatabase` or `Dump` fails here, `SaveLocked` returns before reaching `finalizeSave`, so a database locked by this call stays locked forever (no other instance/user can save until it's unlocked manually).

This is pre-existing behavior from the old `Save`/`SaveAs` callback chain, not introduced by this refactor. But `SaveLocked` is now clean to fix it, e.g. a `defer` at the top:

```go
if !wasLocked {
    defer func() {
        if err := s.Unlock(ctx, database.ID); err != nil {
            s.logError(ctx, fmt.Errorf("unlock database failed: %v", err))
        }
    }()
}
```

(and drop the `!wasLocked` branch from `finalizeSave`).

radnov (commented):
Overall looks good 👍  Noticed one potential issue with lock/unlock during Saving and added as a comment.

pkg/database/service.go
tonsV2: Good catch, fixed. The unlock now runs deferred from SaveLocked so every failure path releases it, and finalizeSave keeps only the unconditional unlock before the row delete plus the re-lock for the wasLocked case. Added TestSaveLockedUnlocksOnDumpFailure covering the dump-failure path.

radnov (approved):

PR #1598, opened 2026-07-20
refactor: extract pkg/kube from instance kubernetes.go

tonsV2:
First step of the stack-operations model (roadmap step 1). The Kubernetes client and its operations move out of pkg/instance into a new leaf package pkg/kube so stack definitions can reference operation implementations without an import cycle (pkg/instance imports pkg/stack). Behavior-preserving move: kubernetesService becomes kube.Client with exported fields and methods, restart is split into RestartDeployment/RestartStatefulSet with the KubernetesResource switch relocated into instance.Service.Restart, and DecryptYaml plus the cluster-resource and ingress/cert discovery helpers move along. pkg/group now depends on pkg/kube for FindResources instead of pkg/instance. A temporary NewKubernetesService bridge keeps the database pod-exec wiring and filestore callers compiling and is removed in the follow-up component operations PR. No route, request, or behavior changes; swagger only updates the x-go-package of ClusterResources to reflect the move.


pkg/kube/workloads.go
```diff
@@ -0,0 +1,196 @@
...
+	if err != nil {
+		return err
+	}
+
+	statefulSetsItems := statefulSetsList.Items
+	if len(statefulSetsItems) == 0 {
+		return fmt.Errorf("no deployment found using the selector: %q", selector)
+	}
+	if len(statefulSetsItems) > 1 {
+		return fmt.Errorf("multiple deployments found using the selector: %q", selector)
```
radnov: Maybe change "deployments" to "stateful sets" here?

radnov (approved):

pkg/kube/workloads.go
tonsV2: Good catch, done in ac47957c. The stateful set arm had inherited the deployment wording from the original combined restart; both the not-found and multiple-found errors now say "stateful set".

PR #1599, opened 2026-07-20
test: retry minio seed exec to fix flaky filestore backup test

radnov:
`TestInstanceHandler/FilestoreBackupMinioViaExec` flakes because `AssertPodIsReady` returns on minio's k8s `PodReady` condition, which does not guarantee the S3 API on :9000 is accepting connections yet. The one-shot seed `mc cp` then fails with `dial tcp 127.0.0.1:9000: connect: connection refused`.

Wrap the idempotent seed exec in `require.Eventually` so it retries until minio serves, matching the existing `require.Eventually` pattern in this file. Also stop silencing the `mc alias set`/`mc mb` pre-commands so retries log why they failed.


radnov:
Superseded by #1598

PR #1601, opened 2026-07-21
feat: per-stack component operations

tonsV2:
Roadmap step 2 of the stack-operations model, targeting version-3.0.

Each stack definition now declares the components it consists of via a Components field on stack.Stack, and each component knows how to restart its own workload: DeploymentComponent patches a Deployment, StatefulSetComponent a StatefulSet. instance.Service.Restart dispatches through the instance's components (all of them when no component is named via ?selector=, else the named one) instead of switching on the KubernetesResource enum, which is deleted. The hardcoded stack-name to PVC-label map becomes per-component PVCPatterns collected by DestroyInstance and passed to the new kube.DeletePVCs, and the kube file holding these reusable operation mechanics is renamed from workloads.go to operations.go. The temporary NewKubernetesService bridge from the pkg/kube extraction is removed; callers use kube.NewClient directly.

im-job-runner deliberately declares no components: it is an old jobs experiment that only labels pods, and jobs get their own design in a separate task; its restart keeps returning an error, now with a clear message.

Only intended behavior changes: restart now works for the combined dhis2 stack (restarts core plus db); unqualified restart targets by im-type instead of im-default, which is equivalent for all current helmfiles. PVC-deletion parity with the old map is asserted by a test, as is that every component name matches an im-type label in its stack's helmfile. pause/resume/status/logs are unchanged and there are no route, request, or swagger changes.

The reviewed design doc (docs/component-operations.md) is committed separately. Follow-up PR: component listing endpoint and per-replica restart.


pkg/instance/service.go
```diff
@@ -595,42 +606,50 @@ func (s Service) Resume(ctx context.Context, instance *model.DeploymentInstance)
...
 		return err
 	}
 
-	instanceStack, err := s.stackService.Find(instance.StackName)
+	components, err := s.stackService.Components(instance.StackName)
 	if err != nil {
 		return err
 	}
+	if len(components) == 0 {
+		return fmt.Errorf("stack %q has no components to restart", instance.StackName)
```
radnov: Should we use `NewBadRequest` or `NewConflict` here, instead of returning a general 500?

pkg/instance/service.go
```diff
@@ -595,42 +606,50 @@ func (s Service) Resume(ctx context.Context, instance *model.DeploymentInstance)
...
-	case stack.StatefulSetResource:
-		return ks.RestartStatefulSet(instance, typeSelector)
-	case stack.DeploymentResource:
-		return ks.RestartDeployment(instance, typeSelector)
-	default:
-		return fmt.Errorf("kubernetes resource not supported: %s", instanceStack.KubernetesResource)
+	if componentName == "" {
+		var errs error
+		for _, component := range components {
+			errs = errors.Join(errs, component.Restart(ctx, client, instance))
```
radnov: A joined error from these Restart calls will result in an error like "something went wrong. We'll look into it if you send us the request id ...". Naming the component at least puts it in the logs, so it's clear which one failed. Something like this:

```suggestion
if err := component.Restart(ctx, client, instance); err != nil {
  errs = errors.Join(errs, fmt.Errorf("restarting component %q: %w", component.ComponentName(), err))
}
```

radnov (approved):

pkg/instance/service.go
tonsV2: Agreed, done in 6c3c8d85. Went with NewBadRequest over NewConflict: in our taxonomy Conflict marks transient state worth retrying (instance starting up, database locked), while a stack having no components is a permanent capability gap, so 400 signals not to retry.

pkg/instance/service.go
tonsV2: Applied verbatim in 6c3c8d85. The %w also keeps the underlying error classifiable through errors.Join, so a typed error from a single component still maps to the right status instead of collapsing to 500.

PR #1602, opened 2026-07-21
refactor: move Stack out of pkg/model into pkg/stack

tonsV2:
Pure relocation, no behavior change. Stack and its definitional companions (StackParameter(s), ParameterProvider(s), RequireCompanionFunc, KubernetesResource) are static hardcoded definitions that are never persisted, unlike the gorm-backed rows in pkg/model, so they move into pkg/stack where they belong.

The stack handler's response DTOs were also named Stack and StackParameter; they become StackResponse and StackParameterResponse. Their swagger:model names (Stack, StackParameter) are kept via annotations, so the API contract is unchanged; the only swagger diff is two cosmetic x-go-name metadata lines.

This is the standalone move that unblocks giving stack.Stack a Components field, so each stack definition can carry its components inline (reworking the name-keyed registry from the component-operations work into fields).

Note: local integration runs hit external-registry network flakes (helm chart CDN / Docker Hub DNS timeouts) unrelated to this structural change; the hermetic suites (stack, database, deployment, storage) and build/vet/swagger are clean. CI runs the integration suite in a cleaner network environment.

### DIFF
--- a/pkg/database/database_integration_test.go
+++ b/pkg/database/database_integration_test.go
@@ -20,6 +20,7 @@ import (
 	"github.com/dhis2-sre/im-manager/pkg/instance"
 	"github.com/dhis2-sre/im-manager/pkg/inttest"
 	"github.com/dhis2-sre/im-manager/pkg/model"
+	"github.com/dhis2-sre/im-manager/pkg/stack"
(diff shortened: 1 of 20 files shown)


radnov (approved):
LGTM 👍

PR #1603, opened 2026-07-22
docs: plan to migrate IM control plane from EKS to a Hetzner VM

radnov:
Planning + docs only (no application code changes) to move the IM control plane off the EKS `instance-cluster` (DEVOPS-684, EKS in extended support) onto Hetzner VMs running docker compose. Opening as a draft for approach review before any build work.

## Approach
- **Two decoupled Hetzner VMs** in `nbg1`: prod on its own `cpx22`, dev + feature envs on a `cpx32`. Isolating prod from the churny, credential-bearing feature envs (arbitrary PR code) rides the existing prod/nonprod credential split. Decoupled from the `k3s-prod` workload cluster (failure isolation, stable resources, independent upgrades).
- Per-env isolated stacks (`prod`, `dev`, `feat-*`): each with its own IM + Postgres + RabbitMQ + Redis. Only Traefik (ingress) and an observability shipper are shared, one set per VM.
- Keep AWS `S3` (DB blobs), `KMS` (SOPS), `SES` (mail); swap IRSA for one scoped static IAM key per VM (prod key on the prod VM, nonprod key on the dev+feat VM).
- Drop `ingress-nginx` / `cert-manager` / `cluster-autoscaler` / `EBS CSI` / `Velero`; replace with Traefik+LE, size-for-peak, docker volumes, and a `pg_dump`->S3 backup cron.
- Monitoring stays on the existing k3s IM Grafana (it is not on EKS); Route53 records repointed per env to the prod / dev+feat VM.
- Cutover via a rehearsed `pg_dump` / `pg_restore` in a short maintenance window; running DHIS2 instances are unaffected throughout.

## Contents
- `docs/runbooks/im-eks-to-vm-spec.md` - target architecture + per-component migrate/drop/replace mapping
- `docs/runbooks/im-eks-to-vm-runbook.md` - cutover runbook (pre-flight gates, dump/restore, rollback)

## Decisions worth a second opinion
- Decouple onto dedicated VMs vs. running the control plane on the existing `k3s-prod` cluster (spec argues decouple).
- Split prod onto its own VM vs. one shared VM for all envs (spec argues split, for prod isolation + the prod/nonprod boundary).
- Feature-env CI: SSH-from-GitHub-Actions `im-envctl` (recommended) vs. a deploy-agent vs. keeping feature envs on k3s.
- Backups: `pg_dump`->S3 with daily/weekly retention instead of Velero.
- The `default` cluster should point at the `k3s-prod` kubeconfig (never nil, which would mean in-cluster config that does not exist on a VM).

### DIFF
--- a/docs/runbooks/im-eks-to-vm-runbook.md
+++ b/docs/runbooks/im-eks-to-vm-runbook.md
@@ -1,0 +1,311 @@
+# Runbook: Migrate the IM control plane from EKS to a Hetzner VM
+
+Relocate the Instance Manager (IM) control plane off the AWS EKS cluster (`DEVOPS-684`, EKS in
+extended support) onto **two Hetzner VMs** running docker compose — prod on its own VM, dev + feature
+environments on a second VM — then decommission EKS. See
+[`im-eks-to-vm-spec.md`](./im-eks-to-vm-spec.md) for the target architecture; this runbook is the
+cutover procedure.
+
+Approach: **rehearsed `pg_dump` / `pg_restore` in a short maintenance window.** No streaming
+replication, no master/slave. Instance workloads already run on Hetzner and are not touched.
+
+**Two-VM target:** `prod` migrates to the **prod VM** (`cpx22`); `dev` migrates to the **dev+feat VM**
+(`cpx32`); per-PR `feat-*` envs are created fresh by CI and are **not** migrated. The steps below
+describe the **prod cutover** to the prod VM — run the same procedure for `dev` against the dev+feat VM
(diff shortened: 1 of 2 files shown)


radnov:
Superseded by dhis2-sre/dhis2-infrastructure#284. The runbook moved to `services/im-vm/RUNBOOK.md` next to the Terraform it drives, and the spec is dropped: the architecture it described is now the Terraform itself, and the reasoning that is worth keeping is in `services/im-vm/README.md`. Nothing here lands in im-manager.

PR #1620, opened 2026-07-27
feat: component listing endpoint and per-replica restart

tonsV2:
Roadmap step 3, targeting version-3.0. Adds GET /instances/:id/components returning each component of an instance with its supported operations and live replicas, so clients can render per-component actions and per-replica restart buttons. The restart endpoint gains a replica query parameter which deletes a single pod after validating it belongs to the component given by selector, letting the controller recreate it; replica without selector is a 400. Components can now declare capabilities beyond the base restart/restartReplica operations, optionally gated on decrypted instance parameters. dhis2-core's core component advertises filestoreBackup without a predicate since all storage backends (minio, filesystem, s3) support it since #1562; the first genuinely parameter-gated capability arrives with CNPG-backed stacks. Concrete component types are named for the chart/technology they operate on (BitnamiPostgresComponent, MinioComponent, DHIS2CoreComponent, ...) and live in pkg/stack next to the stack definitions; the Kubernetes workload kind is an implementation detail of each Restart, and the generic DeploymentComponent/StatefulSetComponent types are gone. chap-db is typed CNPGPostgresComponent, keeping today's statefulset restart behavior until restart is reimplemented via the Cluster CR in the operator step. A future dhis2v2 stack swaps the postgres component type in its definition alone to gain CNPG-specific operations such as native S3 backup. Context is threaded through the kube operation helpers (RestartDeployment, RestartStatefulSet, Pause, Resume, scale), replacing the context.TODO() calls carried over from the verbatim extraction, so request cancellation reaches the Kubernetes API. Covered by unit tests on the component model, a table test asserting each technology type patches the workload kind its chart deploys, and an integration test exercising the listing, the validation error and a real replica restart against k3s.

### DIFF
--- a/pkg/instance/service.go
+++ b/pkg/instance/service.go
@@ -614,7 +614,7 @@ func (s Service) Resume(ctx context.Context, instance *model.DeploymentInstance)
 	return ks.Resume(instance)
 }
 
-func (s Service) Restart(ctx context.Context, instance *model.DeploymentInstance, componentName string) error {
+func (s Service) Restart(ctx context.Context, instance *model.DeploymentInstance, componentName, podName string) error {
 	group, err := s.groupService.Find(ctx, instance.GroupName)
 	if err != nil {
 		return err
@@ -634,6 +634,9 @@ func (s Service) Restart(ctx context.Context, instance *model.DeploymentInstance
 	}
 
 	if componentName == "" {
+		if podName != "" {
+			return errdef.NewBadRequest("restarting a replica requires a component selector")
+		}
 		var errs error
 		for _, component := range components {
 			if err := component.Restart(ctx, client, instance); err != nil {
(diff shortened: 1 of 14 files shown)


pkg/instance/service.go
```diff
@@ -634,6 +634,9 @@ func (s Service) Restart(ctx context.Context, instance *model.DeploymentInstance
 	}
 
 	if componentName == "" {
+		if podName != "" {
+			return errdef.NewBadRequest("restarting a replica requires a component selector")
```
radnov: Aren't we already returning this error in the handler here https://github.com/dhis2-sre/im-manager/pull/1620/changes#diff-0a77e42443c6231c5241d66235af8a4e3b08dfe25a653f17991b4b02922df648R647?

pkg/instance/service.go
```diff
@@ -647,9 +650,51 @@ func (s Service) Restart(ctx context.Context, instance *model.DeploymentInstance
...
 }
 
+type ComponentStatus struct {
+	Name                string           `json:"name"`
+	SupportedOperations []kube.Operation `json:"supportedOperations"`
+	Replicas            []kube.Replica   `json:"replicas"`
+}
+
+// Components lists the instance's components with their supported operations and live replicas.
+// The instance must be decrypted since capability predicates evaluate real parameter values.
```
radnov: Wouldn't it be better to move the decryption step in this service method, instead of relying on the callers passing a decrypted instance? (like here https://github.com/dhis2-sre/im-manager/pull/1620/changes#diff-0a77e42443c6231c5241d66235af8a4e3b08dfe25a653f17991b4b02922df648R689)

radnov (approved):

pkg/instance/service.go
tonsV2: Good catch, it was checked in both places. Kept it once at the top of the service in 19f96c6d since the service owns the contract for any future caller, and dropped the handler check; the service error still maps to a 400.

pkg/instance/service.go
tonsV2: Agreed, done in 19f96c6d. Components now looks up the stack and decrypts the instance itself, and the handler is back to the plain FindDeploymentInstanceById.

PR #1623, opened 2026-07-27
refactor: name components by chart technology

tonsV2:
Stacked on #1620, retargets to version-3.0 when it merges. Re-cuts the concrete component types by chart/technology per design discussion: a component is a BitnamiPostgresComponent or MinioComponent, not a StatefulSetComponent; the workload kind is an implementation detail inside each two-line Restart delegating to the shared helpers. DeploymentComponent/StatefulSetComponent are deleted. chap-db becomes CNPGPostgresComponent, keeping its current statefulset restart no-op behavior against CNPG bare pods until roadmap step 5 implements restart via the Cluster CR. A future dhis2v2 stack swaps the postgres component type in its definition alone to gain CNPG-specific operations. Also threads context through the kube operation helpers (RestartDeployment, RestartStatefulSet, Pause, Resume, scale), removing the context.TODO() calls carried over from the verbatim extraction, so request cancellation propagates to the Kubernetes API. No behavior change otherwise; restart delegation is covered by a new table test over every technology type.

### DIFF
--- a/pkg/instance/service.go
+++ b/pkg/instance/service.go
@@ -597,7 +597,7 @@ func (s Service) Pause(ctx context.Context, instance *model.DeploymentInstance) 
 		return err
 	}
 
-	return ks.Pause(instance)
+	return ks.Pause(ctx, instance)
 }
 
 func (s Service) Resume(ctx context.Context, instance *model.DeploymentInstance) error {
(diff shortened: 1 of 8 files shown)


tonsV2:
Folded into #1620 for a single coherent review.

PR #1628, opened 2026-07-28
feat: component presence predicates and the dhis2-v2 stack

tonsV2:
Roadmap step 9's IM half, targeting version-3.0.

Components can declare a presence predicate evaluated against decrypted instance parameters: Restart now decrypts internally like Components and both filter by presence, so an absent component is a 404 and restart-all skips it. DestroyInstance deliberately stays unfiltered since PVC deletion skips empty selectors and a changed storage type must not orphan the previous component's volumes.

The first consumer is the new dhis2-v2 stack: one release of the dhis2 chart (dhis2-sre/dhis2-core-chart#86) bundling core, a CloudNativePG PostgreSQL cluster and an optional MinIO, replacing the dhis2-db + dhis2-core + minio companion composition. STORAGE_TYPE maps to the chart's storage.type and minio.enabled, database credentials are set once via global.postgresql.auth, hostnames resolve to the chart's services (the CNPG read-write service for the database), and each component carries its own im-type label (dhis2, db, minio).

Seeding uses the chart's marker-guarded post-install job with the seed script adapted to its superuser PG* environment, replacing the Bitnami initdb mechanism.

The db component is a CNPGPostgresComponent with its volume selected via the cnpg.io/cluster label; its CR-based restart lands with roadmap step 5, until then restart targets nothing, unchanged from chap-db today. The minio component only exists when STORAGE_TYPE is minio, so the components endpoint never lists a phantom minio. DATABASE_VERSION is dropped since the PostgreSQL version rides the chart's PostGIS image.

Deploying requires the published dhis2 chart, now released as dhis2 1.0.0 (dhis2-sre/dhis2-core-chart#86 merged with charts/core removed and the dhis2 chart promoted to the only chart); the CHART_VERSION default points at 1.0.0. Integration coverage for a live dhis2-v2 deploy needs the CNPG operator in the test k3s and follows separately.

### DIFF
--- a/pkg/stack/components_test.go
+++ b/pkg/stack/components_test.go
@@ -20,10 +20,16 @@ import (
 // allStacks are the deployable stack definitions; every one must declare its components. The
 // job-runner is exempt: it is an old jobs experiment that only labels pods, and jobs get their
 // own design in a separate task.
-var allStacks = []Stack{
-	DHIS2DB, MINIO, DHIS2Core, DHIS2, DHIS2V2, PgAdmin, WhoamiGo,
-	ChapDB, ChapValkey, ChapWorker, ChapCore,
-}
+var allStacks = func() []Stack {
+	var stacks []Stack
+	for _, s := range All {
+		if s.Name == IMJobRunner.Name {
+			continue
+		}
+		stacks = append(stacks, s)
+	}
+	return stacks
+}()
 
 func TestEveryStackHasUniqueNamedComponents(t *testing.T) {
 	for _, s := range allStacks {
--- a/pkg/stack/components_test.go
+++ b/pkg/stack/components_test.go
@@ -21,7 +21,7 @@ import (
 // job-runner is exempt: it is an old jobs experiment that only labels pods, and jobs get their
 // own design in a separate task.
 var allStacks = []Stack{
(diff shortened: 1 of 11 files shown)


pkg/stack/components_test.go
```diff
@@ -21,7 +21,7 @@ import (
 // job-runner is exempt: it is an old jobs experiment that only labels pods, and jobs get their
 // own design in a separate task.
 var allStacks = []Stack{
-	DHIS2DB, MINIO, DHIS2Core, DHIS2, PgAdmin, WhoamiGo,
+	DHIS2DB, MINIO, DHIS2Core, DHIS2, DHIS2V2, PgAdmin, WhoamiGo,
```
radnov: We also declare the stacks in `main.go`. I think we should have them in one place and reused, so that tests can't draft.

radnov (approved):

pkg/stack/components_test.go
tonsV2: Agreed, fixed in the latest commit: stack.All is the single list, main.go registers stack.All... and both test files derive from it (the components test filters out the job-runner by name, keeping its exemption explicit).

PR #1631, opened 2026-07-29
feat: cnpg component restart through the cluster resource

tonsV2:
First slice of roadmap step 5. The CNPG postgres component's restart stamps the standard kubectl.kubernetes.io/restartedAt annotation on the CloudNativePG Cluster resource through a new dynamic client on kube.Client, the same mechanism as the cnpg kubectl plugin's restart command, so the operator performs a supervised rollout. Previously the component carried the generic statefulset patch, which found nothing against CNPG's bare operator-managed pods; that gap was confirmed during the dhis2-v2 live validation. The Cluster name comes from a ClusterPattern on the component, set for dhis2-v2 (%s-dhis2-postgresql) and chap-db (%s-chap-db-postgres), which makes chap-db's restart operation real for the first time too. RBAC already covers the patch verb on clusters.postgresql.cnpg.io from #1629. Covered by a unit test patching a Cluster through the fake dynamic client; live verification on the version-3-0 environment follows the next env redeploy.

### DIFF
--- a/pkg/kube/client.go
+++ b/pkg/kube/client.go
@@ -20,9 +20,10 @@ import (
 )
 
 // Client talks to a single cluster's Kubernetes API. Construct it with NewClient so RestConfig is
-// populated; the fields are exported so tests can inject a fake Clientset.
+// populated; the fields are exported so tests can inject a fake Clientset and a fake Dynamic client.
 type Client struct {
 	Clientset  kubernetes.Interface
+	Dynamic    dynamic.Interface
 	RestConfig *rest.Config
 }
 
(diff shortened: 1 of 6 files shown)


tonsV2:
Live-tested on this PR's deploy environment (feat-cnpg-restart): deployed a dhis2-v2 instance, then PUT /instances/1/restart?selector=db returned 202, the kubectl.kubernetes.io/restartedAt annotation appeared on the Cluster resource, CNPG reported "Primary instance is being restarted without a switchover", recreated the pod with a new uid on the same volume, and the cluster returned to "Cluster in healthy state". On the current version-3.0 build the same call fails since the statefulset patch finds nothing against operator-managed pods.

radnov (approved):

PR #1654, opened 2026-08-02
feat: push component status over the event stream

tonsV2:
Pod informer events become transient component-status events on the existing notification stream, scoped to the owning instance group: one event per meaningful replica transition (phase, readiness, restarts), with a deleted flag on removals. The initial informer sync and no-op updates are filtered out, and pods of instances IM no longer knows are skipped.

PublishTransient streams without persisting a notification and without an owner, so component status reaches every group member live but never appears in the notification bell. The SSE endpoint also gains a 30 second heartbeat comment frame so idle connections survive proxies.

Stacked on #1652, merge that first. The frontend listener that consumes these events and retires the Refresh button follows separately.

### DIFF
--- a/pkg/instance/componentStatusWatcher.go
+++ b/pkg/instance/componentStatusWatcher.go
@@ -2,6 +2,7 @@ package instance
 
 import (
 	"context"
+	"fmt"
 	"log/slog"
 	"strconv"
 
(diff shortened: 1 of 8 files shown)


pkg/instance/componentStatusWatcher.go
```diff
@@ -0,0 +1,102 @@
...
+	publisher transientPublisher
+}
+
+func NewComponentStatusWatcher(logger *slog.Logger, instances instanceLookup, publisher transientPublisher) *ComponentStatusWatcher {
+	return &ComponentStatusWatcher{logger: logger, instances: instances, publisher: publisher}
+}
+
+var _ cache.ResourceEventHandler = &ComponentStatusWatcher{}
+
+func (w *ComponentStatusWatcher) OnAdd(obj interface{}, isInInitialList bool) {
```
tonsV2: `interface{}`? At least use `any` if a concrete type isn't available. Same goes for multiple other places here

pkg/instance/componentStatusWatcher.go
```diff
@@ -0,0 +1,102 @@
...
+	// The initial list is the informer syncing what already runs; clients load that state through
+	// the components endpoint, so only genuine additions are pushed.
+	if isInInitialList {
+		return
+	}
+	w.publish(obj, false)
+}
+
+func (w *ComponentStatusWatcher) OnUpdate(oldObj, newObj interface{}) {
+	oldPod, okOld := oldObj.(*v1.Pod)
```
tonsV2: Since the type can be anything we should assert this conversion so we avoid panic

pkg/instance/componentStatusWatcher.go
```diff
@@ -0,0 +1,102 @@
...
+
+func (w *ComponentStatusWatcher) OnDelete(obj interface{}) {
+	if unknown, ok := obj.(cache.DeletedFinalStateUnknown); ok {
+		obj = unknown.Obj
+	}
+	w.publish(obj, true)
+}
+
+func (w *ComponentStatusWatcher) publish(obj interface{}, deleted bool) {
+	pod, ok := obj.(*v1.Pod)
```
tonsV2, line 88: Assert conversion?

PR #1666, opened 2026-08-04
feat: add daily check for expiring TLS certificates

tonsV2:
climate.im.dhis2.org served an expired certificate this morning. The cert was not failing to renew, it had nothing renewing it: the Certificate resource was gone and only the Secret remained, which nginx kept serving until it expired.

Instances in a group share one hostname and one TLS secret to stay within Let's Encrypt rate limits, and cert-manager's ingress-shim lets only one ingress own the shared Certificate via an ownerReference. When that instance was destroyed the Certificate was garbage collected with its ingress, and the remaining six ingresses never recreate it because deleting a Certificate does not re-enqueue them. Busy groups like play and android hide this by recreating the Certificate on every deploy. Climate had no deploys, so it aged out.

This adds scripts/util/find-expiring-certificates.py and a daily workflow alongside the existing orphaned-helm-releases check, posting to team-devops-internal. It reports certs served by an Ingress that are expiring or expired, that have no Certificate resource behind them (flagged regardless of days left, since nothing is renewing them), or whose Secret was never created. Scope is limited to certs an Ingress actually serves and cert-manager is expected to manage, so dead leftovers and manually managed secrets stay quiet.

Run against both clusters it found two stuck issuances nobody was watching: api.im-dev.dhis2.org on hetzner and whoami.im-feat.dhis2.org on EKS, both with solver pods running for about a day. Use --ignore-namespace to mute churn-heavy namespaces if the dev noise is unwelcome.

Keeping the shared-secret design as is. This was the first occurrence in four years, so detection is the cheaper fix.

### DIFF
--- a/.github/workflows/find-expiring-certificates.yaml
+++ b/.github/workflows/find-expiring-certificates.yaml
@@ -1,0 +1,71 @@
+name: Find Expiring Certificates
+
+on:
+  schedule:
+    - cron: '30 8 * * *'
+  workflow_dispatch:
+    inputs:
+      days:
+        description: 'Expiry threshold in days'
+        required: false
+        default: '14'
+
+jobs:
+  find-expiring-certificates:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+
+      - name: Install kubectl
+        uses: alexellis/arkade-get@master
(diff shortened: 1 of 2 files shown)


Philip-Larsen-Donnelly (approved):
Seems reasonable

PR #1677, opened 2026-08-05
test: bound the heap of every core the integration tests deploy

radnov:
`TestInstanceHandler` loses processes to the OOM killer, and the symptom lands wherever it likes: `FilestoreBackupMinioViaExec` failed with `backup failed: command terminated with exit code 137` while backing up a single 15-byte object, and an earlier run failed two unrelated `UpdateDeploymentInstance*` subtests in 0.03s each.

Nothing in a deployed core bounds its memory. The stack's `JAVA_OPTS` default is a single space, so no `-Xmx` is ever passed, and the chart sets `resources.requests.memory` with no limit. The JVM therefore finds no cgroup limit and applies `MaxRAMPercentage`, a quarter of whatever `/proc/meminfo` reports, which inside the gnomock k3s container is the whole runner. On a live core pod, where the same absence of limits applies, that is:

```
MaxHeapSize = 8216641536   # 7.65 GiB, 25% of a ~30 GB node
```

`createDHIS2CoreInstance` now passes `-Xmx1g`, prepended so a caller can still override it, which bounds every core the suite deploys in one place.

The bound is asserted rather than assumed, and the first run of this PR earned that: the assertion originally read `/proc/1/cmdline`, on the assumption that catalina expands `JAVA_OPTS` onto the java command line. It failed, while reporting a 1 GiB ceiling sourced from `{command line}` on the very next line. The chart appends `javaOpts` to `JAVA_TOOL_OPTIONS`:

```
- name: JAVA_TOOL_OPTIONS
  value: >-
    -Djava.io.tmpdir=/tmp/dhis ... -Xmx1g
```

Every JVM in the container honours that, and none of them shows it on argv, so a throwaway JVM's flags are the readable surface. The assertion now checks both the value and that it came from `{command line}` rather than `{ergonomic}`, which is what separates a bound that was set from one derived from the node.

It runs in `FilestoreBackupFilesystemViaExec`, which already has a running core pod, so it costs one exec rather than another deploy.

Whether it clears the flake is for CI to say, since the failure rate is roughly one run in three.

#1676 should land first: five of the last seven master failures were the Deploy step, not the tests.

### DIFF
--- a/pkg/instance/instance_helper_test.go
+++ b/pkg/instance/instance_helper_test.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"strconv"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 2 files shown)


Philip-Larsen-Donnelly (approved):
Sounds reasonable, though I don't understand 100% of the description 😬

